PR #12, opened 2023-08-08
Allow optional dependencies

fxmarty:
As per title

### DIFF
--- a/optimum_benchmark/import_utils.py
+++ b/optimum_benchmark/import_utils.py
@@ -1,0 +1,18 @@
+import importlib.util
+
+_torch_available = importlib.util.find_spec("torch") is not None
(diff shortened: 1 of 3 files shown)


optimum_benchmark/import_utils.py
```diff
@@ -0,0 +1,18 @@
+import importlib.util
+
+_torch_available = importlib.util.find_spec("torch") is not None
+_onnxruntime_available = importlib.util.find_spec("onnxruntime") is not None
+_is_openvino_available = importlib.util.find_spec("openvino") is not None
+_is_neural_compressor_available = importlib.util.find_spec("neural-compressor") is not None
```
IlyasMoutawwakil, line 6: I think it should be "neural_compressor" (that's why its test is failing)

fxmarty: Thanks!

PR #13, opened 2023-08-08
Log git SHA when transformers/accelerate are installed from source from local

fxmarty:
As per title.

e.g.

```yml
hydra:
  job:
    env_set:
      TRANSFORMERS_PATH: /home/fxmarty/transformers
```

can be set in the yml config (or manually).

### DIFF
--- a/optimum_benchmark/main.py
+++ b/optimum_benchmark/main.py
@@ -1,14 +1,14 @@
 import os
 import hydra
 import platform
-from typing import Type
+from typing import Type, Dict
 from logging import getLogger
 from hydra.utils import get_class
-from dataclasses import dataclass, MISSING
-from omegaconf import DictConfig, OmegaConf
+from dataclasses import dataclass, MISSING, field
+from omegaconf import OmegaConf
 from hydra.core.config_store import ConfigStore
 
-
+from .utils import get_git_revision_hash
 from optimum.exporters import TasksManager
 from optimum.version import __version__ as optimum_version
 from transformers import __version__ as transformers_version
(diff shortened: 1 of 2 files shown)


optimum_benchmark/main.py
```diff
@@ -78,11 +77,12 @@ class ExperimentConfig:
     )
 
     # ENVIRONMENT CONFIGURATION
-    environment: DictConfig = DictConfig(
-        {
+    environment: Dict = field(default_factory=lambda: {
             "optimum_version": optimum_version,
             "transformers_version": transformers_version,
+            "transformers_commit": None,
```
IlyasMoutawwakil, line 83: can't we put here ?
`transformers_commit": get_git_revision_hash(os.getenv("TRANSFORMERS_PATH"))`
or use a resolver ?

fxmarty: Unfortunately if the env variable `TRANSFORMERS_PATH` is set in the yml, it is not already visible at this point. An alternative could be to autodetect the repo path and check whether it is a git repo, but it is probably overkill.

IlyasMoutawwakil: ah I see. then a resolver would be best.
I prefer not to set the values in the experiment config during the benchmark because even though the manually saved `hydra_config.yaml` will have those changes, the automatically saved `./hydra/config.yaml` won't.

IlyasMoutawwakil:
Can you resolve the conflicts? I'd like to merge but I can't resolve the conflicts because it's on your fork I guess. 
Also, can we make it log the sha of all the hf libraries since we use main branches 😅 dunno if that's too verbose.

PR #14, opened 2023-08-08
Support device_map for torch backend

fxmarty:
This is useful to benchmark large models on several GPUs.

Configs can include to set the number of devices to use.

```yml
hydra:
  job:
    env_set:
      CUDA_VISIBLE_DEVICES: 0,1
```

### DIFF
--- a/optimum_benchmark/main.py
+++ b/optimum_benchmark/main.py
@@ -70,7 +70,6 @@ class ExperimentConfig:
             "cache_dir": None,
             "force_download": False,
             "local_files_only": False,
-            "use_auth_token": False,
         }
     )
 
--- a/optimum_benchmark/trackers/latency.py
+++ b/optimum_benchmark/trackers/latency.py
@@ -50,7 +50,7 @@ class LatencyTracker:
 class PyTorchLatencyTracker(LatencyTracker):
     def __init__(self, backend):
         super().__init__(backend)
-        if backend.device_map:
+        if backend.config.device_map:
             self.hf_device_map = backend.pretrained_model.hf_device_map
             # This logic will break if anything else than device_map="auto" is used.
             self.start_device = min(self.hf_device_map.values())
--- a/optimum_benchmark/trackers/latency.py
+++ b/optimum_benchmark/trackers/latency.py
@@ -52,23 +52,25 @@ class PyTorchLatencyTracker(LatencyTracker):
         super().__init__(backend)
         if backend.config.device_map:
             self.hf_device_map = backend.pretrained_model.hf_device_map
-            # This logic will break if anything else than device_map="auto" is used.
-            self.start_device = min(self.hf_device_map.values())
             self.end_device = max(self.hf_device_map.values())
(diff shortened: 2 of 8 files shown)


IlyasMoutawwakil:
Thanks for addition

IlyasMoutawwakil:
I guess there are some things that should be updated when `device_map` is used, so that the reported metrics are relevant.
Like memory consumption (it should include all devices) and maybe we need to synchronize all devices  for latency?

fxmarty:
Right, good catch! I'll include that in this PR.

optimum_benchmark/main.py
```diff
@@ -70,7 +70,6 @@ class ExperimentConfig:
             "cache_dir": None,
             "force_download": False,
             "local_files_only": False,
-            "use_auth_token": False,
```
fxmarty, line 73: This argument is deprecated in transformers

IlyasMoutawwakil: yes was it substituted with `token` argument ?

fxmarty: Not necessarily, you can use `huggingface-cli login` and it works.

IlyasMoutawwakil (changes requested):

optimum_benchmark/trackers/latency.py
```diff
@@ -46,3 +46,41 @@ def _cpu_latency(self):
...
+        super().__init__(backend)
+        if backend.config.device_map:
+            self.hf_device_map = backend.pretrained_model.hf_device_map
+            self.end_device = max(self.hf_device_map.values())
+            self.device_indexes = set(self.hf_device_map.values())
+        else:
+            self.hf_device_map = None
+            self.end_device = self.device
+            if self.device.type == "cuda":
+                self.device_indexes = {self.device.index if self.device.index is not None else 0}
```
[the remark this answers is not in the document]
fxmarty: @IlyasMoutawwakil When using `CUDA_VISIBLE_DEVICES=3 python my_torch_script.py`, PyTorch remaps the available device to zero (i.e. `torch.cuda.current_device()` returns 0) so I think it is fine.

optimum_benchmark/utils.py
```diff
@@ -27,7 +27,8 @@ def set_seed(seed: int) -> None:
 
 
 def bytes_to_mega_bytes(bytes: int) -> int:
-    return bytes >> 20
+    # Reference: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
+    return bytes * 1e-6
```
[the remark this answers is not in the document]
fxmarty: No. Should it?

optimum_benchmark/utils.py
```diff
@@ -27,7 +27,8 @@ def set_seed(seed: int) -> None:
 
 
 def bytes_to_mega_bytes(bytes: int) -> int:
```
fxmarty, line 29:
```suggestion
def bytes_to_mega_bytes(bytes: int) -> float:
```

IlyasMoutawwakil (approved):

PR #21, opened 2023-08-10
Fix isolation checks when using `device_map`

fxmarty:
This PR:
* Fixes the isolation checks when `device_map` is used for the pytorch backend
* avoids a call to `torch.cuda.current_device()` which may trigger an error `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method` as it requires to initialize CUDA in a subprocess, which is sensitive to the start method.
* adds a yaml example with `device_map`

### DIFF
--- a/examples/cuda_pytorch_inference_gpt2_2gpu.yaml
+++ b/examples/cuda_pytorch_inference_gpt2_2gpu.yaml
@@ -9,6 +9,8 @@ device: cuda
 
 backend:
   device_map: auto
+  initial_isolation_check: false
+  continous_isolation_check: false
 
 hydra:
   job:
--- a/examples/cuda_pytorch_inference_gpt2_2gpu.yaml
+++ b/examples/cuda_pytorch_inference_gpt2_2gpu.yaml
@@ -1,0 +1,19 @@
+defaults:
+  - base_config # inherits from base config
+  - _self_ # for hydra 1.1 compatibility
+
+experiment_name: cuda_pytorch_inference_gpt2_2gpu
+
+model: gpt2
+device: cuda
+
+backend:
+  device_map: auto
+
+hydra:
+  job:
+    env_set:
+      CUDA_VISIBLE_DEVICES: 0,1
+  sweeper:
+    params:
+      benchmark.input_shapes.batch_size: 1,2,4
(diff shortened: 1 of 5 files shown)


examples/cuda_pytorch_inference_gpt2_2gpu.yaml
```diff
@@ -0,0 +1,19 @@
...
+backend:
+  device_map: auto
+
+hydra:
+  job:
+    env_set:
+      CUDA_VISIBLE_DEVICES: 0,1
+  sweeper:
+    params:
+      benchmark.input_shapes.batch_size: 1,2,4
```
IlyasMoutawwakil, line 19: for now sweeps are not tested (needs the `-m` flag in the test script) so this config will only test the default `batch_size` value.
is it necessary to test multiple batch sizes ?

fxmarty: Let's keep it like that, it's fine if only the first is tested.

I added several because there was an error only when using sweeps before (PID already using the device due to the cuda context not cleaned).

IlyasMoutawwakil:
isolation checks should be disabled because the tests fail when someone is using the DGX at the same time.

fxmarty:
@IlyasMoutawwakil Should be good now!

fxmarty:
@IlyasMoutawwakil Do you think the failing test is related to this PR? Is the workflow running on the merge commit or on the branch commit? I can't reproduce the failing workflow on neither of them.

IlyasMoutawwakil:
I believe the workflow runs on branch commit. there's a lot of inefficiency in it. I'm trying to avoid the image building each time in another PR. I'm also setting them to not be launched automatically, it gets very messy with the queue.

IlyasMoutawwakil:
it seems the error is coming from `llm_int8_threshold=0`. transformers must have removed it in source, I will check.

IlyasMoutawwakil:
Aah I see why this is happening. since we're not using quantization in the config you added, the `llm_int8_threshold=0` is not being filtered from `kwargs` by the `BnBQuantizationConfig`. in these cases I generally use:
```python
**(
    {"llm_int8_threshold": 0}
    if config.load_in_4bit or config.load_in_8bit
    else {}
),
```
but the best way should be passing a `quantization_config`, we could do that when adding GPTQ support.

fxmarty:
@IlyasMoutawwakil It should be fixed with https://github.com/huggingface/optimum-benchmark/pull/21/commits/805839210682c38fbce9f2472d82c79e1ebf51c4! Feel free to merge if it looks good to you!

PR #22, opened 2023-08-10
Add TensorRT execution provider to onnxruntime backend

shuyuelan:
With this change, we can run with TensorRT execution provider in onnxruntime backend with `execution_provider` flag. For example,`optimum-benchmark --config-dir examples --config-name onnxruntime_bert_inference device=cuda execution_provider=Tensorrt`.

### DIFF
--- a/gpu_requirements.txt
+++ b/gpu_requirements.txt
@@ -10,6 +10,7 @@ git+https://github.com/huggingface/diffusers.git
 # dunno why the above installs onnxruntime and not onnxruntime-gpu
 onnxruntime-gpu==1.15.1
 
+tensorrt
 omegaconf==2.3.0
 hydra-core==1.3.2
 hydra_colorlog==1.2.0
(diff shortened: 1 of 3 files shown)


IlyasMoutawwakil:
Hi, that's already supported in this example https://github.com/huggingface/optimum-benchmark/blob/main/examples/tensorrt_onnxruntime_inference_bert.yaml

shuyuelan:
Hi Ilyas. Thanks for sharing the new config files with TRT EP usage. I did not notice it before. What I added here is parameterizing the execution provider for general usage. It is up to you whether to merge this PR. I am ok with both.

IlyasMoutawwakil:
Thank you @shuyuelan, I prefer keeping backend dependent parameters/terminology in the backend. This way they're in the output `hydra_config.yaml` only when it makes sense. You can use the parameter `backend.provider` just like `execution_provider`.

PR #26, opened 2023-08-11
Enable inference profiling

jingyao-zhang:
Using `benchmark.profile=true` either in cmd line or in config files enables profiling during inference


IlyasMoutawwakil:
Hi, I prefer having profiling in a separate benchmark named `profiling`. it was removed to reduce the complexity of the `inference` benchmark and because profiling and inference results are not complementary but rather one is included in the other.

jingyao-zhang:
I see. After separating them, I will PR.

PR #30, opened 2023-08-18
Distributed tests and isolated multirun

IlyasMoutawwakil:
This moves and add support for the distributed tests that were added in https://github.com/huggingface/optimum-benchmark/pull/18 and https://github.com/huggingface/optimum-benchmark/pull/14


optimum_benchmark/backends/base.py
```diff
@@ -49,8 +49,8 @@ class BackendConfig(ABC):
     intra_op_num_threads: Optional[int] = None
 
     # isolation options
-    initial_isolation_check: bool = True
-    continous_isolation_check: bool = True
+    initial_isolation_check: bool = False
+    continous_isolation_check: bool = False
```
fxmarty, line 53: I think this should be left to True by default. It is an useful feature.

IlyasMoutawwakil: okay! I had the impression that it doesn't work with docker containers so I thought it'd make sense to  disable it until it's stable. but it seems it works fine with docker.

fxmarty: Ah yes with docker you need to use `docker run --pid host ...` for it to work. I had the same issue, should have documented it

fxmarty: https://github.com/fxmarty/transformers-regression-test/commit/9531138bd458fe2259a825e3c7d79a539e2abad6

PR #35, opened 2023-08-23
Backend pytorch logging mistake when config.amp_autocast is True.

aoowweenn:
We should make logging for config.amp_dtype when config.amp_autocast is True since self.amp_dtype is not set yet.


IlyasMoutawwakil:
nice catch! I'll merge it rn.

Issue #27, opened 2023-08-14
CUDA_VISIBLE_DEVICES is not captured by torch


fxmarty:
As per title.

For example setting

```
hydra:
  job:
    env_set:
      CUDA_VISIBLE_DEVICES: 4,5
```

does not work and the model will be loaded on devices 0, 1. I believe there's a bug somewhere in the trainer or transformers (or hydra, or torch) where things mess up when `CUDA_VISIBLE_DEVICES` is set after torch import.

Launching with `CUDA_VISIBLE_DEVICES=3,4 optimum-benchmark` solves the issue.

I think we should abandon the hydra/job/env_set approach, this just means that the CI should be edited accordingly.

IlyasMoutawwakil:
I experimented with it yesterday. the problem is that we import `torch` before the experiment start (entry point for hydra).
I'll try some stuff and see if I can get it to work.

IlyasMoutawwakil:
@fxmarty would it make sense to only allow distributed inference/training when `CUDA_VISIBLE_DEVICES` is explicitly set ?

fxmarty:
Yes I think we should do that. That's what I ended up doing in the regression tests, it's slightly ugly but we are sure everything works fine (otherwise there can be nasty bugs as this one): https://github.com/fxmarty/transformers-regression-test/blob/b98f838c8a24fe8d00ee7991bc57420c7050517a/run_benchmark.sh#L16-L23

IlyasMoutawwakil:
solved in https://github.com/huggingface/optimum-benchmark/pull/36

PR #28, opened 2023-08-14
Add warmup tracking for training benchmark

fxmarty:
For AMD, we'll need to separate warmup from the rest of the training (see here: https://github.com/huggingface/transformers/compare/main...ROCmSoftwarePlatform:transformers:master), which can be fairly different, see e.g.

```
2023-08-14T11:28:16.5447387Z   0%|          | 0/375 [00:00<?, ?it/s]
2023-08-14T11:28:16.6619759Z   0%|          | 1/375 [00:01<08:34,  1.38s/it]
2023-08-14T11:28:16.7760017Z   1%|          | 3/375 [00:01<02:28,  2.50it/s]
2023-08-14T11:28:16.8847029Z   1%|▏         | 5/375 [00:01<01:22,  4.47it/s]
2023-08-14T11:28:16.9943622Z   2%|▏         | 7/375 [00:01<00:55,  6.58it/s]
2023-08-14T11:28:17.1037176Z   2%|▏         | 9/375 [00:01<00:42,  8.67it/s]
2023-08-14T11:28:17.2145885Z   3%|▎         | 11/375 [00:01<00:34, 10.62it/s]
2023-08-14T11:28:17.3276767Z   3%|▎         | 13/375 [00:02<00:29, 12.31it/s]
2023-08-14T11:28:17.4366979Z   4%|▍         | 15/375 [00:02<00:26, 13.66it/s]
2023-08-14T11:28:17.5456206Z   5%|▍         | 17/375 [00:02<00:24, 14.85it/s]
2023-08-14T11:28:17.6545471Z   5%|▌         | 19/375 [00:02<00:22, 15.79it/s]
2023-08-14T11:28:17.7639035Z   6%|▌         | 21/375 [00:02<00:21, 16.49it/s]
2023-08-14T11:28:17.8718203Z   6%|▌         | 23/375 [00:02<00:20, 17.00it/s]
2023-08-14T11:28:17.9815874Z   7%|▋         | 25/375 [00:02<00:20, 17.45it/s]
2023-08-14T11:28:18.0905970Z   7%|▋         | 27/375 [00:02<00:19, 17.67it/s]
2023-08-14T11:28:18.2000539Z   8%|▊         | 29/375 [00:02<00:19, 17.86it/s]
2023-08-14T11:28:18.3093587Z   8%|▊         | 31/375 [00:03<00:19, 17.99it/s]
2023-08-14T11:28:18.4188164Z   9%|▉         | 33/375 [00:03<00:18, 18.08it/s]
[...]

PR #36, opened 2023-08-24
Major refactoring

IlyasMoutawwakil:
This refactoring's purpose is to regain separation of concerns and removing patches. 
This will allow for better unittests ; testing trackers separately from backends, and generators separately from benchmarks.
Another reason for this refactoring is to standardize the interaction between `BackendConfig`<->`Backend`, `BenchmarkConfig`<->`Benchmark`, `Benchmark`<->`Backend` and interactions within a backend (optimization, quantization, etc.)
Some key points:
- many of the arguments that end with `config` are now ~optional~ *empty dictionaries* since their usage is generally related to another control argument. With this, `hydra_config.yaml` will not contain things that are not used by the benchmark script. For example, ort optimization/quantization configs, `torch.compile`'s config, etc.
- Each backend being dependent on many third party packages, it's always better to keep standard library imports separated from third party imports and from local repo imports.
- Minimizing `pytorch` usage outside of `PytorchBackend` and backends that require pytorch (solves https://github.com/huggingface/optimum-benchmark/issues/27)
- DDP control arguments were moved to the `pytorch` backend for consistency.
- As much as possible of the processing of hydra configs should happen in its `__post_init__`. Any other processing should not affect the types/values of the config

Some benefits of this refactoring:
- All training benchmarks benefit from the `warmup_steps` through the measurement callback previously implemented for DDP.
- solves https://github.com/huggingface/optimum-benchmark/issues/27.
- Deprecation warnings for some arguments that will be removed soon in favor of more flexible arguments: `generate_config` instead of `new_tokens`, `forward_config` instead of `num_images_per_prompt` or any other argument that can be passed to diffusion pipelines.

Only breaking change is DDP.


optimum_benchmark/backends/base.py
```diff
@@ -1,55 +1,60 @@
...
 import gc
 
 
-import torch
-from torch import Tensor
-from datasets import Dataset
+import shutil
 from psutil import cpu_count
-from omegaconf import DictConfig
+from diffusers import DiffusionPipeline
```
fxmarty, line 12: Does this make diffusers a hard requirement of optimum-benchmark?

optimum_benchmark/backends/base.py
```diff
@@ -119,7 +134,9 @@ def check_initial_isolation(self) -> None:
             cuda_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
             if cuda_devices is None:
                 LOGGER.warning(
-                    "Asked to check the initial device isolation, but the variable CUDA_VISIBLE_DEVICES was not set. Defaulting to checking on the first device."
+                    "Asked to check the initial device isolation, "
+                    "but the variable CUDA_VISIBLE_DEVICES was not set. "
+                    "Defaulting to checking on the first device."
```
fxmarty, line 139: Are you using ruff or anything? I think it can make sense to add a workflow to check the style (like in optimum & transformers), and a Makefile to make style.

optimum_benchmark/backends/base.py
```diff
@@ -150,87 +169,60 @@ def check_continuous_isolation(self) -> None:
...
         raise NotImplementedError("Backend must implement train method")
 
     def delete_pretrained_model(self) -> None:
-        if hasattr(self, "pretrained_model"):
+        try:
             del self.pretrained_model
-            gc.collect()
+        except AttributeError:
+            # benchmark might fail before the model is loaded
+            pass
```
fxmarty, line 209: It could be worth logging the error here?

optimum_benchmark/backends/onnxruntime.py
```diff
@@ -1,39 +1,54 @@
...
 
-from optimum_benchmark.backends.base import Backend, BackendConfig
-from optimum_benchmark.backends.utils import main_export, randomize_weights
-from optimum_benchmark.profilers.ort_profiler import ORTProfilingWrapper
-from optimum_benchmark.utils import infer_device_id
 
 OmegaConf.register_new_resolver(
     "is_gpu",
-    lambda device: torch.device(device).type == "cuda",
+    lambda device: "cuda" in device.lower() or "tensorrt" in device.lower(),
```
fxmarty, line 51: device is used to select the EP?

optimum_benchmark/backends/base.py
```diff
@@ -62,18 +67,28 @@ class BackendConfig(ABC):
...
+    pretrained_model: Union[PreTrainedModel, DiffusionPipeline]
     pretrained_processor: Optional[PreTrainedProcessor]
+    pretrained_config: Optional[PretrainedConfig]
 
-    def __init__(self, model: str, task: str, device: str, hub_kwargs: DictConfig):
+    def __init__(self, model: str, task: str, device: str, hub_kwargs: Dict[str, Any]):
         self.model = model
         self.task = task
-        self.device = torch.device(device)
+        self.device = device
```
fxmarty, line 91: why?

optimum_benchmark/backends/onnxruntime.py
```diff
@@ -70,103 +81,109 @@ class ORTConfig(BackendConfig):
...
+    config: ORTConfig
+
     def __init__(
-        self, model: str, task: str, device: str, hub_kwargs: DictConfig
+        self, model: str, task: str, device: str, hub_kwargs: Dict[str, Any]
     ) -> None:
         super().__init__(model, task, device, hub_kwargs)
-
-        from optimum.pipelines import ORT_SUPPORTED_TASKS
+        self.device = torch.device(device)
```
fxmarty, line 186: Can't it mess up if `device` is `tensorrt` (reading the resolver above)?

optimum_benchmark/backends/pytorch.py
```diff
@@ -38,32 +46,25 @@
 @dataclass
 class PyTorchConfig(BackendConfig):
     name: str = "pytorch"
-    version: str = torch.__version__
+    version: str = torch_version
```
fxmarty, line 49: What is the point?

optimum_benchmark/backends/pytorch.py
```diff
@@ -72,10 +73,88 @@ class PyTorchConfig(BackendConfig):
     disable_grad: bool = "${is_inference:${benchmark.name}}"  # type: ignore
     eval_mode: bool = "${is_inference:${benchmark.name}}"  # type: ignore
 
+    # training options
+    use_ddp: bool = False
+    ddp_config: Optional[Dict[str, Any]] = None
+
+    def __post_init__(self):
```
fxmarty, line 80: really clean!

Does this non-greedy approach means that if defaults of optimum-benchmark are changed, a previous benchmark config yml may not be accurately loaded?

optimum_benchmark/backends/pytorch.py
```diff
@@ -72,10 +73,88 @@ class PyTorchConfig(BackendConfig):
...
+        if self.torch_compile:
+            self.torch_compile_kwargs = OmegaConf.merge(
+                self.torch_compile_kwargs
+                if self.torch_compile_kwargs is not None
+                else {},
+                DEFAULT_COMPILE_CONFIG,
+            )
+
+        if self.device_map is not None:
+            assert self.device_map in ["auto", "sequential"], (
```
fxmarty, line 97: I think it can make sense to use `ValueError` here (and below as well) for readability.

optimum_benchmark/backends/pytorch.py
```diff
@@ -72,10 +73,88 @@ class PyTorchConfig(BackendConfig):
...
+            assert (
+                CUDA_VISIBLE_DEVICES is not None
+            ), "Pytorch DDP training benchmark requires CUDA_VISIBLE_DEVICES to be set."
+        else:
+            self.ddp_config = None
+
 
 class PyTorchBackend(Backend):
+    name: str = "pytorch"
+    config: PyTorchConfig
```
fxmarty, line 153: Why is config a class attribute and not an instance attribute? It is probably fine since we always have a single instance of `PyTorchBackend`, but it can be bug prone otherwise.

optimum_benchmark/backends/pytorch.py
```diff
@@ -86,179 +165,172 @@ def configure(self, config: PyTorchConfig) -> None:
...
-                # Diffusers does not support device_map being a torch.device, thus if not provided, move to device here.
-                self.pretrained_model.to(self.device)
-
-    def prepare_for_profiling(
-        self,
-        input_names: List[str],
-        input_shapes: Dict[str, int],
-    ) -> None:
+    def prepare_for_profiling(self, input_names: List[str]) -> None:
         LOGGER.info("Preparing model for profiling")
```
fxmarty, line 333: It could make sense to support pytorch profiler as well in the future (https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html) that does not rely on torch.fx so is easier to use (but serves a different purpose)

optimum_benchmark/backends/pytorch.py
```diff
@@ -268,128 +340,75 @@ def prepare_for_profiling(
...
-            # TODO: support multi-node training. Hydra is probably not the good infra for that though.
-            if training_config.ddp_config.max_nodes != 1:
-                raise ValueError("PyTorch DDP training benchmark currently supports only training on a single node.")
-                
-            launch_config = LaunchConfig(**training_config.ddp_config)
-            LOGGER.info(f"PyTorch DDP launch config: {launch_config}")
-            
-            # TODO: The backend instance can not be passed here (cannot pickle 'weakref' object) so the nn.Module is passed directly.
-            # It is not clear who is using weakref though.
+    @record
```
fxmarty, line 363: What is this doing?

optimum_benchmark/benchmarks/training.py
```diff
@@ -46,193 +45,58 @@ class TrainingConfig(BenchmarkConfig):
...
-            self.training_metrics = {
-                "training_throughput": training_output.metrics[
-                    "train_samples_per_second"
-                ],
-                "train_runtime": training_output.metrics["train_runtime"],
-            }
+        self.training_metrics = {
+            # warmup metrics
+            "warmup_runtime": trainer_state.warmup_runtime,
+            "warmup_throughput()": trainer_state.warmup_samples_per_second,
```
fxmarty, line 92:
```suggestion
            "warmup_throughput": trainer_state.warmup_samples_per_second,
```

optimum_benchmark/benchmarks/training.py
```diff
@@ -46,193 +45,58 @@ class TrainingConfig(BenchmarkConfig):
...
+            # training metrics
+            "train_runtime": trainer_state.train_runtime,
+            "training_throughput": trainer_state.train_samples_per_second,
+            # overall training metrics
+            "overall_train_runtime": trainer_state.overall_train_runtime,
+            "overall_training_throughput": trainer_state.overall_train_samples_per_second,
```
fxmarty, line 98: should probably align on train or training

optimum_benchmark/benchmarks/training.py
```diff
@@ -46,193 +45,58 @@ class TrainingConfig(BenchmarkConfig):
...
-                # local_addr: Optional[str] = None
-            }
-            if self.ddp_config is not None:
-                ddp_config.update(self.ddp_config)
-            self.ddp_config = ddp_config
-
 
 class TrainingBenchmark(Benchmark):
+    name: str = "training"
+    config: TrainingConfig
```
fxmarty, line 63: same comment

optimum_benchmark/experiment.py
```diff
@@ -114,13 +113,11 @@ class ExperimentConfig:
 
 @hydra.main(version_base=None)
 def run_experiment(experiment: DictConfig) -> None:
-    # By default, Hydra populates the metadata object_type with the ones from ExperimentConfig but the object_type should really be
-    # one of the subclass (e.g. PyTorchBackendConfig instead of BackendConfig). This is required to call `to_object`.
-    experiment = remap_to_correct_metadata(experiment)
+    from omegaconf import SCMode
 
-    # This is required to trigger __post_init__. Reference: https://github.com/omry/omegaconf/issues/377
```
fxmarty, line 121: I think this comment is useful to know why this code is there.

optimum_benchmark/experiment.py
```diff
@@ -114,13 +113,11 @@ class ExperimentConfig:
...
-    experiment = OmegaConf.to_object(experiment)
-    experiment = OmegaConf.create(experiment)
+    experiment = OmegaConf.to_container(
+        experiment, structured_config_mode=SCMode.INSTANTIATE
+    )
```
fxmarty, line 120: `resolve` is False by default for `to_container` while it is True for `to_object`. Does this indeed resolves? Why and where?

I am not sure to remember why I put `OmegaConf.create`, why is it not necessary anymore?

optimum_benchmark/utils.py
```diff
@@ -148,7 +139,8 @@ def check_only_this_process_is_running_on_cuda_device(
                 ]
             )
 
-            # check if there is a process running on device_id that is not the current process
+            # check if there is a process running on
+            # device_id that is not the current process
```
fxmarty, line 143: I think PEP-8 80 characters per line is too strict. It could make sense to add a config like https://github.com/huggingface/optimum/blob/main/pyproject.toml#L19-L23

tests/configs/base_config.yaml
```diff
@@ -18,14 +17,6 @@ hydra:
...
     # this is useful for saving outputs in a separate directory
     chdir: true
-  launcher:
-    # we set the number of jobs to 2 since when using 1, joblib reuses the same process
-    n_jobs: 2
-    prefer: processes
-    backend: multiprocessing
-  sweeper:
-    # now we force the sweeper to run one job at a time, achieving almost perfect isolation
-    max_batch_size: 1
```
fxmarty, line 28: Why remove it in the end?

tests/configs/distributed_cuda_pytorch_training_bert_ddp.yaml
```diff
@@ -4,15 +4,16 @@ defaults:
...
 device: cuda
 
-benchmark:
+backend:
   use_ddp: true
+
+benchmark:
   dataset_shapes:
-    dataset_size: 1200
+    dataset_size: 120
```
fxmarty, line 16: Doesn't this break the minimum warmup steps?

optimum_benchmark/benchmarks/training.py
```diff
@@ -1,31 +1,30 @@
...
-OmegaConf.register_new_resolver("device_count", lambda: torch.cuda.device_count())
 
 
 @dataclass
 class TrainingConfig(BenchmarkConfig):
     name: str = "training"
     _target_: str = "optimum_benchmark.benchmarks.training.TrainingBenchmark"
 
+    # training options
+    warmup_steps: int = 2
```
fxmarty, line 26: This is much too low. Can we leave a higher default? Also, changing defaults like this is breaking older config that are not greedy (not saving all params), I think it can be an issue.

See https://github.com/huggingface/optimum-benchmark/pull/28

fxmarty (commented):
Great work! I think it is indeed nicer to dissociate as much as possible the backends, and to use `__post_init__` when possible. Still a bit worried for the possible change of defaults (like in transformers: https://github.com/huggingface/transformers/pull/25237#issuecomment-1691254241), but I'm not sure if it is an issue or not.

I did not review everything but left a few comments!

optimum_benchmark/backends/base.py
IlyasMoutawwakil: I'll use `transformers` pipeline instead

optimum_benchmark/backends/onnxruntime.py
IlyasMoutawwakil: that's a mistake, I forgot that I already decided not to go with this design in the past

optimum_benchmark/backends/pytorch.py
IlyasMoutawwakil: batter logging for pytorch multiprocessing

optimum_benchmark/benchmarks/training.py
IlyasMoutawwakil: I was thinking of adding metrics like in inference

optimum_benchmark/benchmarks/training.py
IlyasMoutawwakil: Yes

optimum_benchmark/benchmarks/training.py
IlyasMoutawwakil: depends on the hardware but we can keep it 40.

optimum_benchmark/backends/base.py
IlyasMoutawwakil: in an attempt to reduce torch usage outside of its backend. 
I'll revert it since I already found a way to not import torch before `CUDA_VISIBLE_DEVICES` is set.

optimum_benchmark/backends/pytorch.py
fxmarty: Do you have an example?

optimum_benchmark/benchmarks/training.py
fxmarty: Yes agree, it probably depend on the hardware / model / batch size / etc.

tests/configs/base_config.yaml
IlyasMoutawwakil: I was getting some errors with ddp so I thought `joblib` might be interfering with `torch.multiprocessing`.
Turns out it was because the other backends were also imported (nncf) which altered the behavior of pytorch (used some unpicklable objects, specifically a lambda function). 
In the future it might be worth moving the model instantiation (for training benchlark) in the training worker to avoid any problems between pickle and modeling code.

optimum_benchmark/backends/base.py
fxmarty: makes sense

optimum_benchmark/utils.py
IlyasMoutawwakil: done !

optimum_benchmark/backends/pytorch.py
IlyasMoutawwakil: now it's handled by resolvers and `importlib` to avoid importing torch.

optimum_benchmark/backends/pytorch.py
IlyasMoutawwakil: it means that they'll rely on the defaults of `optimum-benchmark`. this won't be much of a problem for things like experimenting with the speedup of an optimization once, but it is a problem for regression tests, in which case it's better to keep using the same version of the repo, unless interested by a new feature, in which case the user will need to check everything is still the same. I guess we can add a warning for future changes whenever we're planning one, like `disable_xxx_fusion` in `optimum` optimizations.

optimum_benchmark/backends/pytorch.py
IlyasMoutawwakil: thanks, I changed the errors to be more meaningful.

optimum_benchmark/benchmarks/training.py
IlyasMoutawwakil: I added some better typing (still typing noob 😅).

optimum_benchmark/experiment.py
IlyasMoutawwakil: yes true, `to_object` is safer and more strict, some things can go unnoticed with `to_container` and stay as strings instead of being resolved

IlyasMoutawwakil: I think there's no benefit to using `to_create` because with it we just go back to `DictConfig` objects. 
not sure why we had it there, I thought it was part of the `__post_init__` patch.

optimum_benchmark/backends/pytorch.py
IlyasMoutawwakil: it is an instance attribute without a default value. the only class attribute is the name.

optimum_benchmark/experiment.py
IlyasMoutawwakil: true, I put it back.

IlyasMoutawwakil:
@fxmarty I added a tag `before-refactoring` that can be used in case this creates any issues.

PR #37, opened 2023-08-28
Re: Major refactoring #36

IlyasMoutawwakil:
https://github.com/huggingface/optimum-benchmark/pull/36 didn't fully solve the `CUDA_VISIBLE_DEVICES` https://github.com/huggingface/optimum-benchmark/issues/27 issue because I ended up relying on torch for device handling and third party packages imported it anyways.
This should do it. I will also add a test later that checks `CUDA_VISIBLE_DEVICES` is respected when passed in config.
Edit: At the end it was mostly `optimum.exporters` that imported torch when `TasksManager` is imported.

### DIFF
--- a/optimum_benchmark/backends/base.py
+++ b/optimum_benchmark/backends/base.py
@@ -3,7 +3,6 @@ import os
 import random
 import shutil
 from abc import ABC
-from dataclasses import dataclass
 from logging import getLogger
 from multiprocessing import Process
 from typing import (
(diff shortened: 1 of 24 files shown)


IlyasMoutawwakil:
I found out my GPU and ORT Training tests (exit code) are not propagated to the workflow 😅 probably because they're launched through a script or docker. I'll fix them in another PR. 
I was launching them locally each time before pushing so I never noticed until now.

PR #41, opened 2023-08-28
OnnxRuntime DDP Training

IlyasMoutawwakil:
This PR adds DDP support to `onnxruntime` backend by reusing components from `pytorch` backend.
Also adds test for this scenario.


optimum_benchmark/backends/pytorch/backned.py
```diff
@@ -1,5 +1,4 @@
 import gc
-import os
```
fxmarty, line 2: backned.py?

fxmarty: What does backned means?

PR #42, opened 2023-08-29
CodeCarbon emissions support

IlyasMoutawwakil:
This PR adds support for emissions tracking in the `inference` benchmark through [CodeCarbon](https://github.com/mlco2/codecarbon).


optimum_benchmark/trackers/emissions.py
```diff
@@ -0,0 +1,24 @@
...
+
+class EmissionsTracker:
+    def __init__(self):
+        self.emissions: float = 0.0
+
+    @contextmanager
+    def track(self, interval=1, file_prefix=""):
+        emission_tracker = CodeCarbonEmissionsTracker(
+            measure_power_secs=interval,
+            output_file=f"{file_prefix}_emissions.csv",
```
regisss, line 17: So we can retrieve any measure provided by CodeCarbon in this CSV file right?

optimum_benchmark/benchmarks/inference/benchmark.py
```diff
@@ -77,16 +79,24 @@ def run_forward_tracking(self, backend: "Backend") -> None:
...
+            with memory_tracker.track(interval=self.forward_latency / 10):
                 _ = backend.forward(forward_input)
 
             self.forward_peak_memory = memory_tracker.get_peak_memory()
             LOGGER.info(f"\t+ Forward pass peak memory: {self.forward_peak_memory} (MB)")
 
+        if self.config.emissions:
+            LOGGER.info("\t+ Tracking forward pass emissions")
+            emissions_tracker = EmissionsTracker()
+            with emissions_tracker.track(interval=self.forward_latency / 10, file_prefix="forward"):
```
regisss, line 91: So we should have more or less 10 measures during the forward pass right?

optimum_benchmark/benchmarks/inference/benchmark.py
```diff
@@ -77,16 +79,24 @@ def run_forward_tracking(self, backend: "Backend") -> None:
...
+                _ = backend.forward(forward_input)
+
+            self.forward_emissions = extract_three_significant_digits(emissions_tracker.get_emissions())
+            LOGGER.info(f"\t+ Forward pass emissions: {self.forward_emissions} (kgCO2)")
+            LOGGER.info("\t+ A detailed report is available at the benchmark directory")
```
regisss, line 96: If the benchmark directory is accessible from here, I would add it to the log with a f-string

regisss (commented):
Nice PR! I left several comments to make sure I understand a few things :slightly_smiling_face:

optimum_benchmark/trackers/emissions.py
IlyasMoutawwakil: Yes exactly. All the CodecCarbon measures are logged primarily in this file. The tracker only returns the carbon emissions but it needs internet to infer it from energy and location. Should we use the offline tracker instead ?

optimum_benchmark/benchmarks/inference/benchmark.py
IlyasMoutawwakil: Yes just as rule of thumb I use the same value for memory measures. It should be noted that the refresh rate of power measures coming from `pynvml` is much slower.

IlyasMoutawwakil: it seems the first option won't work because CodeCarbon uses a thread which interferes with inference (due to the GIL).

optimum_benchmark/benchmarks/inference/benchmark.py
IlyasMoutawwakil: done

optimum_benchmark/trackers/emissions.py
regisss: Does it automatically fall back to the offline tracker if there is no internet connection?

optimum_benchmark/benchmarks/inference/benchmark.py
regisss: I would definitely not measure latency and energy in the same forward pass. So the 2nd solution seems better to me.

optimum_benchmark/benchmarks/inference/benchmark.py
```diff
@@ -77,16 +83,33 @@ def run_forward_tracking(self, backend: "Backend") -> None:
...
                 _ = backend.forward(forward_input)
 
             self.forward_peak_memory = memory_tracker.get_peak_memory()
             LOGGER.info(f"\t+ Forward pass peak memory: {self.forward_peak_memory} (MB)")
 
+        if self.config.energy:
+            LOGGER.info("\t+ Tracking forward pass energy consumption")
+            num_forward_passes = 0
+            energy_tracker = EnergyTracker()
+            with energy_tracker.track(interval=1, file_prefix="forward"):
```
regisss, line 96: With pynvml we cannot get a better frequency than every 1 second?

optimum_benchmark/trackers/emissions.py
IlyasMoutawwakil: They rather implement two trackers, one is [online](https://mlco2.github.io/codecarbon/usage.html#online-mode) which does some api calls to get info about carbon emission rates and one is [offline](https://mlco2.github.io/codecarbon/usage.html#offline-mode) with a non-default `country_iso_code` argument.

optimum_benchmark/benchmarks/inference/benchmark.py
IlyasMoutawwakil: We can get higher measure rates but after experimenting with different values it seems that doesn't change much of the end result. My intuition is that the GPU is already in a steady-state after the warmup and latency tracking (so power doesn't fluctuate a lot) and energy is just `average_power * time`.
Should we use higher frequency nonetheless ?

optimum_benchmark/trackers/emissions.py
regisss: Okay I see. What happens if we use the online tracker without internet access? An error is raised? If yes, we can catch it in a try block and run the offline tracker in the except one.

optimum_benchmark/benchmarks/inference/benchmark.py
regisss: Sounds good, let's keep it that way then :+1: 
Anyway, we'll have opportunities to test various frequencies.

optimum_benchmark/trackers/emissions.py
IlyasMoutawwakil: what about the `country_iso_code` argument ? I don't think you can get location without internet access

regisss: Ah sorry I missed that bit. So the user has to provide this `country_iso_code` right?

IlyasMoutawwakil: yes, should we use something like an environment variable to handle it ? instead of having it in the benchmark config.

regisss: Yes why not

PR #57, opened 2023-09-19
Add RoCm support

fxmarty:
As per title.

There may be more to do:
* raise error with bitsandbytes (as not supported), warning with BT (as flash/mem-efficient are not supported on RoCm), etc.

I'll test more before merging!


optimum_benchmark/trackers/memory.py
```diff
@@ -39,19 +40,43 @@ def get_peak_memory(self):
...
+            for device_index in self.device_ids:
+                handle = nvml.nvmlDeviceGetHandleByIndex(device_index)
+                handles.append(handle)
+            yield
+            for handle in handles:
+                meminfo = nvml.nvmlDeviceGetMemoryInfo(handle)
+                self.peak_memory += meminfo.used
+
+            nvml.nvmlShutdown()
+        elif is_rocm_system():
```
IlyasMoutawwakil, line 62: thismethod is only called when `self.device.type == cuda`. Either we add a method `_rocm_peak_memory` or rename this one to `_gpu_peak_memory` and add `rocm` in the condition above.

IlyasMoutawwakil: hmm unless torch sees `rocm` as `cuda` 🤔, is that the case ?

IlyasMoutawwakil: yeah it is the case 😅 https://pytorch.org/docs/stable/notes/hip.html

IlyasMoutawwakil:
@fxmarty can you add this to the setup file

```python
try:
    subprocess.run(["rocm-smi"], stdout=subprocess.DEVNULL)
    INSTALL_REQUIRES.append("pyrsmi@git+https://github.com/RadeonOpenCompute/pyrsmi.git")
except FileNotFoundError:
    pass
```

PR #87, opened 2023-11-22
Full support tensor parallelism with Deepspeed

IlyasMoutawwakil:
This PR adds support for tensor parallelism by using `deepspeed` and `torchrun`.

It comes with some refactoring to introduce a new set of objects called `launchers`, which allow us to customize how experiments are launched.


IlyasMoutawwakil:
Another thing that is changed in this PR is how we do isolation checks. AMD GPUs can't be perfectly isolated from one another, so a process running on GPU 0 will affect all other GPUs, so what we do is we create a `torch.distributed.TCPStore` where we put all the PIDs that are permitted on our GPUs. For now we use as much isolation processes as worker process, but we can reduce that to only one.

Issue #95, opened 2023-12-05
More tests


IlyasMoutawwakil:
Most of these features are covered in `examples` but not in the testing suite.
- Quantization schemes: 
  - Onnxruntime @aliabdelkader 
  - GPTQ @lopozz 
  - AWQ @aliabdelkader 
  - BnB @aliabdelkader 
- Optimizations:
  - FAv2
  - torch.compile. @aliabdelkader

IlyasMoutawwakil:
@aliabdelkader yess I left these as first issues for anyone interested in contributing and having these settings tested as part of the CI. You just need to add a configuration and make sure that it's tested in the CI (might require adding some packages).

aliabdelkader:
#take 
ok great, I am working on it.

lopozz:
Hi @aliabdelkader and @IlyasMoutawwakil I benchmarked some GPTQ models in the past week and I'd be up to write a test for this quantization scheme if you're not already on it. Let me know

IlyasMoutawwakil:
thanks ! would love to review the quantization configs ! make sure u include `backend.no_weights: true,false` when possible to ur sweep configs as the no weights feature is probably one of the most important features of `optimum-benchmark` and making sure it's always compatible with quantization schemes seems interesting

aliabdelkader:
@lopozz no i did not work on that scheme yet. So, I would not mind if you want to write the test for it. 
@IlyasMoutawwakil sure, I will include the `backend.no_weights: true,false` sweep to the quantizatio test configs.

lopozz:
Hi @IlyasMoutawwakil I was trying the same configuration file of #125 and I got this error 
```
AttributeError: 'LlamaRotaryEmbedding' object has no attribute 'cos_cached'
```
I believe it is caused by the newest refactoring. The last commit I see is  97e0b9e18a09ca3552f7d771cdfd3fa134c55a16 (more makefile stuff) of 4 days ago, so I believe I am working with the latest version. Should I open a new issue for this?

IlyasMoutawwakil:
@lopozz I think it's a transformers issue https://github.com/unslothai/unsloth/issues/168#issuecomment-1943749354, try downgrading ur transformers version

aliabdelkader:
Hi,

I just want to mention that I wont be able to work on the point related to FAv2 in case someone else wants to pick it up.

PR #141, opened 2024-02-29
Added test llama-2-7b with GPTQ quant. scheme

lopozz:
test TheBloke/Llama-2-7B-GPTQ with pytorch backend and cuda hardware. #95


IlyasMoutawwakil:
Thanks a lot for the PR ! I have few suggestions:
- let's use `TheBloke/TinyLlama-1.1B-Chat-v0.3-GPTQ` to reduce download time like in [trt-llm](https://github.com/huggingface/optimum-benchmark/blob/main/tests/configs/tensorrt_llm.yaml#L14) (this is just the smallest llama I can think of)
- let's use the compose api, ie inherit stuff from other configs instead of rewriting them like [here](https://github.com/huggingface/optimum-benchmark/blob/main/tests/configs/cpu_inference_pytorch_gpt_sweep.yaml).

lopozz:
Hi thanks for the feedback. Should I create a default file for llama like it was done for _gpt_ or _timm_? Also I am currently working with a machine with one GPU. hence, the backend.device_ids: 0

IlyasMoutawwakil:
no need, just reuse the ones already provided as much as possible and explicit the rest:
```yaml
defaults:
  - backend: pytorch # default backend
  # order of inheritance, last one overrides previous ones
  - _base_ # inherits from base config
  - _inference_ # inherits from inference config
  - _cuda_ # inherits from cuda config
  - _self_ # hydra 1.1 compatibility

experiment_name: cuda_inference_pytorch_gptq

backend:
  model: TheBloke/TinyLlama-1.1B-Chat-v0.3-GPTQ
  quantization_config:
    exllama_config:
      version: 2

hydra:
  sweeper:
    params:
      backend.no_weights: true,false
```

lopozz:
Hi @IlyasMoutawwakil I just modifid the file with your feedback. Let me know if I can do something else

IlyasMoutawwakil:
@lopozz thanks, can you also add the pip installation to the test workflow

lopozz:
Thanks for the support @IlyasMoutawwakil, it is my first pull request, but I think I solved the issue. Also, seeing the comment https://github.com/huggingface/optimum-benchmark/pull/144#issuecomment-1981232399 I did not modify the setup.py

Main changes: pip installation of optimum and auto-gptq packages

Let me know if you have further feedback.

IlyasMoutawwakil:
@lopozz thanks a lot for working on this.
`optimum` and `auto-gptq` are still missing from `setup.py`'s extras.
I would suggest not adding `optimum` as a standalone but rather grouping both deps in `auto-gptq`
```python
    "bitsandbytes": ["bitsandbytes"],
    "auto-gptq": ["optimum", "auto-gptq"],
```
This will probably not work with ROCm 5.6 & 5.7 and CUDA 11.8 😅 as for these systems an extra url is required to download the right wheels, as explained in: https://github.com/AutoGPTQ/AutoGPTQ?tab=readme-ov-file#installation
It can be solved using `dependency_links` as explained in https://stackoverflow.com/a/30064248, but tell me if this is too complicated for the scope of your PR.

lopozz:
@IlyasMoutawwakil 
update of the last commit:

- added 3 new identifiers in setup.py for cuda121, cuda118 and rocm to install optimum and auto-gptq according to the corresponding index urls.

I did not used dependency_links as you suggested because deprecated starting with pip version 19.0 (released 2019-01-22) https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html#specifying-dependencies-that-aren-t-in-pypi-via-dependency-links. Let me know if you agree, in case I can modify it.

setup.py
```diff
@@ -70,6 +70,9 @@
     "peft": ["peft"],
     "autoawq": ["autoawq@git+https://github.com/casper-hansen/AutoAWQ.git"],
     "bitsandbytes": ["bitsandbytes"],
+    "auto-gptq-cu118": ["optimum", "auto-gptq@https://huggingface.github.io/autogptq-index/whl/cu118/auto-gptq/auto_gptq-0.7.1%2Bcu118-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"],
+    "auto-gptq-cu121": ["optimum", "auto-gptq"],
+    "auto-gptq-rocm": ["optimum", "auto-gptq@https://huggingface.github.io/autogptq-index/whl/rocm573/auto-gptq/auto_gptq-0.7.1%2Brocm5.7.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"]
 }
```
IlyasMoutawwakil, line 76: this won't work for rocm 5.6, let's just remove it and remove gptq tests on rocm ci by adding `and not gptq` to pytest keywords

Makefile
```diff
@@ -15,8 +15,8 @@ CLI_MISC_REQS := testing
 
 CLI_CUDA_ONNXRUNTIME_REQS := testing,timm,diffusers
 CLI_ROCM_ONNXRUNTIME_REQS := testing,timm,diffusers
-CLI_CUDA_PYTORCH_REQS := testing,timm,diffusers,deepspeed,peft,bitsandbytes,autoawq
-CLI_ROCM_PYTORCH_REQS := testing,timm,diffusers,deepspeed,peft,autoawq
+CLI_CUDA_PYTORCH_REQS := testing,timm,diffusers,deepspeed,peft,bitsandbytes,autoawq,auto-gptq-cu118
+CLI_ROCM_PYTORCH_REQS := testing,timm,diffusers,deepspeed,peft,autoawq,auto-gptq-rocm
```
IlyasMoutawwakil, line 19: no need

.github/workflows/test_cli_rocm_pytorch.yaml
```diff
@@ -51,4 +52,4 @@ jobs:
           --device /dev/dri/renderD129
           --entrypoint /bin/bash
           opt-bench-rocm:${{ matrix.image.rocm_version }}
-          -c "pip install requests && pip install -e .[testing,diffusers,timm,deepspeed,peft,autoawq] && pytest -k 'cli and cuda and pytorch and not bnb' -x"
+          -c "pip install requests && pip install -e .[testing,diffusers,timm,deepspeed,peft,autoawq-rocm] && pytest -k 'cli and cuda and pytorch and not bnb' -x"
```
IlyasMoutawwakil, line 55: let's add `and not gptq`

IlyasMoutawwakil (changes requested):

IlyasMoutawwakil:
Thanks, I left some comments, don't forget to run styling. 
There appear to be a `SIGSEGV` when testing GPTQ, probably something wrong with the cleanup code, I'll investigate the cause of that.

IlyasMoutawwakil:
okay found why, in the code I uses `exllama_version` in `exllama_config` but it's supposed to be just `version`.
will fix that in another PR and merge it quickly.

IlyasMoutawwakil:
fixed in #165

IlyasMoutawwakil:
Thanks for addition @lopozz great work on your first PR 🤗

IlyasMoutawwakil (approved):

PR #172, opened 2024-03-28
Add OpenVINO GPU support

helena-intel:
Add GPU support for OpenVINO.

Tested on desktop with ARC GPU and modified pytorch_bert.yaml example (just changed backend and device):
image

I also plan to add an OpenVINO LLM example.


IlyasMoutawwakil:
Thanks a lot for adding and testing this !
is there any way to measure the VRAM (GPU memory) in this case ?

PR #178, opened 2024-04-10
Trackers revamp

IlyasMoutawwakil:
- Standardizing the latency trackers and making them only track, the timing is handled by a separate class `Timer`.
- Making the memory tracker more robust when used with `launch`, by monitoring the main benchmark process (so no need to aggregate VRAM over multiple processes) and supporting both per process and global VRAM tracking at the same time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,6 @@ run_docker_cuda:
 	--gpus all \
 	--shm-size 64G \
 	--env USE_CUDA="1" \
-	--env PROCESS_SPECIFIC_VRAM="0" \
 	--volume $(PWD):/workspace \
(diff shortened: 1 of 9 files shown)


IlyasMoutawwakil:
@franchukpetro I know you've been using VRAM measurements by setting `PROCESS_SPECIFIC_VRAM`. 
This PR removes that environment variable in favor of reporting both (`memory.max_global_vram` and `memory.max_process_vram`), hope this makes VRAM benchmarking easier 🤗

franchukpetro:
Makes sense, thanks a lot for letting me know, @IlyasMoutawwakil!

PR #192, opened 2024-05-02
Remove unnecessary surrogates attached to double quotes

yamaura:
Hello, thank you for the wonderful library.
I found unnecessary surrogates, so I removed them.


IlyasMoutawwakil:
thanks for spotting it ! 
I thought about where it came from and I think it's my Arabic keyboard which gets enabled by mistake sometimes 😅 "َُِ

PR #218, opened 2024-07-01
Fix `PyTorchBackend` TP vs DP inputs distribution across replicas and shards

### LINKED ISSUE #217: BUG Report: Incorrect latency calculation for inference scenario with Tensor Parallel.
Sample numbers are divided by world size in each GPU process.

optimum_benchmark/scenarios/inference/scenario.py Line 64.

``` python
       if is_torch_distributed_available() and torch.distributed.is_initialized():
            self.logger.info("\t+ Distributing batch size across processes")
            if self.config.input_shapes["batch_size"] % torch.distributed.get_world_size() != 0:
                raise ValueError(
                    "The batch size must be divisible by the number of processes in a distributed environment"
                )
            self.config.input_shapes["batch_size"] //= torch.distributed.get_world_size()
```

This part of code makes sure the total batch size is correct for replica inference mode (launcher=torchrun, disable tensor parallel),   , but for tensor parallel mode total batch size is no longer correct. The actual batch size is divided by world size, which results in incorrect throughput caculation.

optimum_benchmark/scenarios/inference/scenario.py. Line 406.
```python
  def atomic_prefill_volume(self) -> int:  # in tokens
      if {"input_ids", "prompt", "prompts"} & set(self.inputs.keys()):
          # text conditioned generation (1 bos token or sequence_length tokens)
          return self.config.input_shapes["batch_size"] * max(self.config.input_shapes["sequence_length"], 1)
      else:
          # image/audio conditioned generation (1 bos token)
          return self.config.input_shapes["batch_size"]
```

IlyasMoutawwakil:
fixes #217 
@luzengxiangcn

### DIFF
--- a/optimum_benchmark/scenarios/inference/scenario.py
+++ b/optimum_benchmark/scenarios/inference/scenario.py
@@ -5,16 +5,12 @@ from transformers import LogitsProcessorList
 from ...backends.base import Backend, BackendConfigT
 from ...benchmark.report import BenchmarkReport
 from ...generators.input_generator import InputGenerator
-from ...import_utils import is_torch_distributed_available
 from ...task_utils import IMAGE_DIFFUSION_TASKS, TEXT_GENERATION_TASKS
 from ...trackers.energy import Efficiency, EnergyTracker
 from ...trackers.latency import LatencyTracker, PerTokenLatencyLogitsProcessor, Throughput
 from ...trackers.memory import MemoryTracker
 from ..base import Scenario
 from .config import InferenceConfig
-
-if is_torch_distributed_available():
-    import torch.distributed
 
 PER_TOKEN_BACKENDS = ["pytorch", "onnxruntime", "openvino", "neural-compressor"]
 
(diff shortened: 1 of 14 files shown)


optimum_benchmark/scenarios/inference/scenario.py
```diff
@@ -61,14 +57,6 @@ def __init__(self, config: InferenceConfig) -> None:
         super().__init__(config)
 
     def run(self, backend: Backend[BackendConfigT]) -> BenchmarkReport:
-        if is_torch_distributed_available() and torch.distributed.is_initialized():
```
luzengxiangcn, line 64: Great! Works for all TP distribution. not only  pytorchbackend but also vllm and tensorRT.

luzengxiangcn:
Great Job! Thanks very much!

optimum_benchmark/scenarios/inference/scenario.py
IlyasMoutawwakil: Yeah for those the TP happens in the engine (no need for torchrun laucher).

Issue #230, opened 2024-07-19
Process isolation is not working on MPS device


baptistecolle:
Setting up the launcher to `process` on an MPS device causes the benchmark to be stuck permanently on “Sending report to main process.”

Steps to reproduce:
Run this configuration on an MPS device (Mac M-series).
```
defaults:
  - benchmark
  - scenario: inference
  - **launcher: process # launcher: inline works**
  - backend: pytorch
  - _base_
  - _self_

name: pytorch_bert

scenario:
  latency: true
  memory: true
  input_shapes:
    batch_size: 1
    sequence_length: 128

backend:
  device: mps
  no_weights: true
  model: bert-base-uncased
```
(example adapted from `pytorch_bert.yaml`)

Temporary Solution: set to `launcher: inline`

Remarks: `launcher.start_method: fork` does not also work on mac

IlyasMoutawwakil:
now that I think about it, this might be a problem in your shared memory size, did you run this on docker ?

baptistecolle:
No, i installed the project from source and ran it

Issue #117, opened 2024-02-01
What other library that optimum-benchmark support other than transformer


L1-M1ng:
Can I use Optimum-benchmark to evaluate the performance of qwen.cpp or llama.cpp？

L1-M1ng:
I want to evaluate the inference latency, throughput, parameter numbers of a custom llm

IlyasMoutawwakil:
`optimum-benchmark` support `transformers`, `timm` and `diffusers` as part of the `pytorch` backend. And then there's `optimum` and its subpackages like `optimum-intel`, `optimum-nvidia`, etc.
There's also a `text-generation-server` backend which uses `docker-py` to benchmark the server end-to-end (i.e. with communications overhead).
currently benchmarking `llama.cpp` is not supported, how would you suggest implementing it ? might possible with a server client api (like tgi) but I'm not sure.

IlyasMoutawwakil:
@L1-M1ng I would love to review a PR with llama.cpp support, https://github.com/abetlen/llama-cpp-python seems to be the most starred python bindings

IlyasMoutawwakil:
llama cpp support added in #231 🚀

zaidalyafeai:
@IlyasMoutawwakil , I am trying to run `optimum-benchmark --config-dir examples/ --config-name llama_cpp_text_generation` but I got this error 
```
Error getting class at optimum_benchmark.backends.llama_cpp.backend.LlamaCppBackend: Error loading 'optimum_benchmark.backends.llama_cpp.backend.LlamaCppBackend'
```

PR #231, opened 2024-07-19
Add llama.cpp backend

baptistecolle:
### Overview
This PR introduces `llama.cpp` as a backend for the Optimum benchmark (see issue #117).

### Changes
- Added an example in the `examples` folder demonstrating how to run the `llama.cpp` backend:

```yaml
defaults:
  - benchmark
  - scenario: inference
  - launcher: inline
  - backend: llama_cpp
  - _base_
  - _self_

name: llama_cpp_llama

backend:
  device: mps
  model: TheBloke/TinyLlama-1.1B-Chat-v1.0-GGUF
  task: text-generation
  filename: tinyllama-1.1b-chat-v1.0.Q4_0.gguf

scenario:
  input_shapes:
    batch_size: 1
    sequence_length: 256
    vocab_size: 32000
  generate_kwargs:
    max_new_tokens: 100
    min_new_tokens: 100
```

###  Current limitations:
- Benchmarking is limited to a batch size of 1 due to limitations of the `llama-cpp-python` binding.
- MPS process isolation is not working (see issue #230).
- No weights mode has not been implemented

### Performance
The metrics were tested by comparing the results of the benchmark with a PyTorch backend and a llama.cpp backend. The performance results are close, and I can provide the full .json files if needed. Due to their low readability, they are not included here directly.
#### CLI output: (tested on M3 Pro CPU)
**Performance with llama.cpp backend**
Screenshot 2024-07-22 at 14 46 40
**Performance with the pytorch backend**
Screenshot 2024-07-22 at 14 49 42

The performance difference might be due to the significant amount of copying between devices with PyTorch, as shown below:
Screenshot 2024-07-22 at 14 50 50
Furthermore, llama.cpp is optimized for Mac, which could explain the higher performance. Let me know if you want me to investigate the performance difference further


optimum_benchmark/backends/base.py
```diff
@@ -68,7 +68,13 @@ def __init__(self, config: BackendConfigT):
             self.automodel_loader = get_timm_automodel_loader()
             self.pretrained_processor = None
             self.generation_config = None
-
+        elif self.config.library == "llama_cpp":
+            self.logger.info("\t+ Benchmarking a Llama.cpp model")
+            self.pretrained_config = None
+            self.model_shapes = {}
+            self.model_type = self.config.task
```
IlyasMoutawwakil, line 75: I think it's better to use the model type info in https://huggingface.co/TheBloke/TinyLlama-1.1B-Chat-v1.0-GGUF/blob/main/config.json gguf repos usually have this or a proper transformers config https://huggingface.co/second-state/Mistral-Nemo-Instruct-2407-GGUF/blob/main/config.json

optimum_benchmark/backends/llama_cpp/backend.py
```diff
@@ -0,0 +1,93 @@
...
+from .config import LlamaCppConfig
+
+
+class LlamaCppBackend(Backend[LlamaCppConfig]):
+    NAME: str = "llama_cpp"
+
+    def __init__(self, config: LlamaCppConfig) -> None:
+        super().__init__(config)
+
+        self.config.name
```
IlyasMoutawwakil, line 18: remove

optimum_benchmark/backends/llama_cpp/backend.py
```diff
@@ -0,0 +1,93 @@
...
+            self.logger.info("\t+ Loading no weights model")
+            raise NotImplementedError("No weights model is not yet implemented")
+
+        self.validate_task()
+
+        self.logger.info("\t+ Creating backend temporary directory")
+        self.tmpdir = TemporaryDirectory()
+        self.logger.info("\t+ Loading pretrained model")
+        self.load_model_from_pretrained()
+        self.tmpdir.cleanup()
```
IlyasMoutawwakil, line 30: should be moved to `load` because we track the first model loading

optimum_benchmark/backends/llama_cpp/backend.py
```diff
@@ -0,0 +1,93 @@
...
+    def load(self) -> None:
+        self.load_model_from_pretrained()
+
+    def load_model_from_pretrained(self) -> None:
+        """
+        Load the pretrained model from the given model name (normally GGUF, GGML)
+        """
+
+        if self.config.task not in TEXT_GENERATION_TASKS:
+            raise NotImplementedError(f"Llama.cpp does not support task {self.config.task}")
```
IlyasMoutawwakil, line 41: already validated the task at this point

optimum_benchmark/backends/llama_cpp/config.py
```diff
@@ -0,0 +1,35 @@
...
+    _target_: str = "optimum_benchmark.backends.llama_cpp.backend.LlamaCppBackend"
+
+    no_weights: bool = False
+    library: str = "llama_cpp"
+    filename: Optional[str] = None
+
+    def __post_init__(self):
+        super().__post_init__()
+
+        print(self.task)
```
IlyasMoutawwakil, line 27: remove

optimum_benchmark/backends/llama_cpp/config.py
```diff
@@ -0,0 +1,35 @@
...
+
+        if self.task not in TEXT_GENERATION_TASKS:
+            raise NotImplementedError(f"Llama.cpp does not support task {self.task}")
```
IlyasMoutawwakil, line 30: remove.

optimum_benchmark/backends/llama_cpp/config.py
```diff
@@ -0,0 +1,35 @@
...
+
+        self.device = self.device.lower()  # type: ignore
+
+        if self.device not in ["cuda", "mps", "cpu"]:
+            raise ValueError(f"Llama.cpp Backend only supports 'cpu', 'mps' and 'cuda' devices, got {self.device}")
```
IlyasMoutawwakil, line 35: how do you choose which device to target ? I didn't see it anywhere in the code

optimum_benchmark/import_utils.py
```diff
@@ -258,4 +268,6 @@ def get_hf_libs_info():
         "timm_commit": get_git_revision_hash("timm"),
         "peft_version": peft_version() if is_peft_available() else None,
         "peft_commit": get_git_revision_hash("peft"),
+        "llama_cpp_version": llama_cpp_version(),
+        "llama_cpp_commit": get_git_revision_hash("llama_cpp"),
```
IlyasMoutawwakil, line 272: Not an HF lib. We have the backend's version in its config.

optimum_benchmark/task_utils.py
```diff
@@ -75,6 +75,9 @@ def infer_library_from_model_name_or_path(model_name_or_path: str, revision: Opt
         if inferred_library_name is None:
             raise RuntimeError(f"Could not infer library name from repo {model_name_or_path}.")
 
+        if "gguf" in model_name_or_path.lower():
+            inferred_library_name = "llama_cpp"
+
```
IlyasMoutawwakil, line 80: remove. at this point the code will have errored and the hub already tags repos with gguf files https://huggingface.co/models?library=gguf

IlyasMoutawwakil (changes requested):
Thanks a lot ! Very awesome work !
Only missing test configs and GitHub workflows 🤗
Hopefully I fix the process launcher with mps by the time the mps workflows start running

optimum_benchmark/backends/llama_cpp/backend.py
```diff
@@ -0,0 +1,93 @@
...
+        return inputs
+
+    def forward(self, inputs: Dict[str, Any], kwargs: Dict[str, Any]) -> Any:
+        """
+        Forward pass of the model\
+        Evaluate the probabilites of the given tokens
+        """
+        return self.pretrained_model.eval(
+            tokens=inputs["input_ids"],
+        )
```
IlyasMoutawwakil, line 70: forward is never called with a text-generation-model so it makes more sense to use https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#embeddings and add `feature-extraction` task to supported tasks

baptistecolle: @IlyasMoutawwakil Good point, do you prefer I add embeddings to llama.cpp in this PR or should I create an issue and do it later?

optimum_benchmark/backends/llama_cpp/config.py
baptistecolle: Indeed, I forgot to set up the target device. I do not see this as an option for the Python binding or even llama.cpp itself. I believe the device is automatically selected, so what I can do is get this chosen device and save it in the .json as the backend. It seems a bit like a workaround but I have not found a cleaner solution

optimum_benchmark/backends/llama_cpp/backend.py
IlyasMoutawwakil: in this one since it's a very minimal change

optimum_benchmark/backends/llama_cpp/config.py
```diff
@@ -0,0 +1,31 @@
...
+
+    no_weights: bool = False
+    library: str = "llama_cpp"
+    filename: Optional[str] = None
+
+    def __post_init__(self):
+        super().__post_init__()
+
+        # if self.task not in TEXT_GENERATION_TASKS + TEXT_EMBEDDING_TASKS:
+        #     raise NotImplementedError(f"Llama.cpp does not support task {self.task}")
```
regisss, line 26: Should this be uncommented?

optimum_benchmark/backends/llama_cpp/backend.py
```diff
@@ -0,0 +1,93 @@
+from tempfile import TemporaryDirectory
+from typing import Any, Dict, Tuple
+
+from llama_cpp import Llama
+
+from optimum_benchmark.backends.base import Backend
+
+from .config import LlamaCppConfig
```
regisss, line 8:
```suggestion
from ..base import Backend
from .config import LlamaCppConfig
```
Let's not mix absolute and relative imports, it can lead to errors that are pretty annoying to debug. We usually prefer using relative imports in the source code.

regisss (commented):
Very clean PR!
I left a couple comments.

Additionally, since benchmarking for llama.cpp is limited to a batch size of 1, I think it may be good to add a comment right above the batch size field in the 3 example configs. Unless an error/warning is raised somewhere in the code but I didn't see it.

baptistecolle:
> Very clean PR! I left a couple comments.
> 
> Additionally, since benchmarking for llama.cpp is limited to a batch size of 1, I think it may be good to add a comment right above the batch size field in the 3 example configs. Unless an error/warning is raised somewhere in the code but I didn't see it.

https://github.com/huggingface/optimum-benchmark/pull/231/files#r1690885402

optimum_benchmark/task_utils.py
baptistecolle: Thanks for the remark about tags i did not know this feature. I am now using it to setup the inferred_library_name properly.

optimum_benchmark/backends/llama_cpp/backend.py
baptistecolle: I added support for it

optimum_benchmark/backends/base.py
baptistecolle: I change it to now try to load the transformers config directly. The transformers loading default to model type info if the config is not a proper transformer config so both case should be handled now

baptistecolle:
Thanks for the review. I implemented the necessary changes

I added: 
- fix different formatting issue
- structure of the code, absolute import, forgotten debug print statement...)
- added support for embedding models with llama.cpp
- added tests and github workflow for llama.cpp backend

Let me know if you have more remarks

optimum_benchmark/task_utils.py
```diff
@@ -69,6 +69,9 @@ def infer_library_from_model_name_or_path(model_name_or_path: str, revision: Opt
         model_info = huggingface_hub.model_info(model_name_or_path, revision=revision)
         inferred_library_name = getattr(model_info, "library_name", None)
 
+        if "gguf" in model_info.tags:
+            inferred_library_name = "llama_cpp"
+
```
IlyasMoutawwakil, line 74: I think the inferred library is enough in this case, no need to make `llama_cpp` the default for repos that have a `gguf` tag (a lot of official repos have both `safetensors` and `gguf` weights), maybe if `inferred_library_name` is still `None` at the end we can check the tags for `gguf`, otherwise the default should stay `transformers` for backward compatibility.

.github/workflows/test_cli_llama_cpp.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - "tests/**"
+      - "setup.py"
+
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+
+jobs:
+  run_cli_llama_cpp_tests:
+    runs-on: ubuntu-latest
```
IlyasMoutawwakil, line 30: I think to run on mps you'll have to use the macos-latest, if it works that is 😅.
If it doesn't (like with pytorch mps) then let's just annotate the tests as cpu tests.

baptistecolle: let's wait to make it MPS (due to https://github.com/actions/runner-images/issues/9918). I added them as CPU in the meantime

optimum_benchmark/task_utils.py
baptistecolle: Good point. So I changed it and now defined the library inside the config instead of inferring it. I also modified the infer_model_type_from_model_name_or_path and infer_task_from_model_name_or_path so that you can supply a library if it isdefined. This prevents them from inferring a different library that the one submitted in the config.

tests/configs/_llama_cpp_.yaml
```diff
@@ -0,0 +1,4 @@
+backend:
```
IlyasMoutawwakil, line 1: this sub config is not used anywhere

optimum_benchmark/task_utils.py
```diff
@@ -168,8 +175,11 @@ def infer_task_from_model_name_or_path(model_name_or_path: str, revision: Option
     return inferred_task_name
 
 
-def infer_model_type_from_model_name_or_path(model_name_or_path: str, revision: Optional[str] = None) -> str:
-    library_name = infer_library_from_model_name_or_path(model_name_or_path, revision=revision)
+def infer_model_type_from_model_name_or_path(
+    model_name_or_path: str, revision: Optional[str] = None, library_name: Optional[str] = None
+) -> str:
+    if library_name is None:
+        library_name = infer_library_from_model_name_or_path(model_name_or_path, revision=revision)
```
IlyasMoutawwakil, line 182: why ?

optimum_benchmark/backends/base.py
```diff
@@ -68,7 +68,12 @@ def __init__(self, config: BackendConfigT):
             self.automodel_loader = get_timm_automodel_loader()
             self.pretrained_processor = None
             self.generation_config = None
-
+        elif self.config.library == "llama_cpp":
+            self.logger.info("\t+ Benchmarking a Llama.cpp model")
+            self.pretrained_config = get_transformers_generation_config(self.config.model, **self.config.model_kwargs)
```
IlyasMoutawwakil, line 73:
```suggestion
            self.generation_config = get_transformers_generation_config(self.config.model, **self.config.model_kwargs)
```

IlyasMoutawwakil: if this what you meant

examples/llama_tiny_mps.yaml
```diff
@@ -0,0 +1,23 @@
+defaults:
```
IlyasMoutawwakil, line 1: there's already a llama example..

IlyasMoutawwakil:
one or two examples are enough, there's a lot of repetition there

optimum_benchmark/backends/base.py
IlyasMoutawwakil: I don't think the generation config is used anywhere by the current implementation of llama cpp

baptistecolle: Indeed, it is not needed, so i removed it

baptistecolle:
Indeed, I created multiple config during development and forgot to remove it from the pr. I fixed it now

IlyasMoutawwakil (approved):

PR #234, opened 2024-07-23
MPS support

IlyasMoutawwakil:
Attempting the same level of support for pytorch MPS as pytorch CPU


IlyasMoutawwakil:
https://github.com/actions/runner-images/issues/9918

PR #238, opened 2024-08-01
Add t4 for llm perf leaderboard

baptistecolle:
## Summary

This PR adds Nvidia T4 to the LLM-Perf Leaderboard:

### Fixes
- Fixed the `trust_remote_code` issue that was broken for the CI/CD pipeline.

### Features
- Added a new machine configuration (T4) to run the LLM performance benchmarking code.
- Updated the A100 configuration to use the new `runs-on` definition (without tags).

### Related PRs and Discussions
- [huggingface/optimum-benchmark#239](https://github.com/huggingface/optimum-benchmark/pull/239)
- [Discussion #30](https://huggingface.co/spaces/optimum/llm-perf-leaderboard/discussions/30)

### Workflow Trigger Changes
One thing that I am unsure about is that I modified the trigger for the workflow. To reduce unnecessary compute, you can manually trigger the workflow, and it is triggered with each new release of the repo ([releases](https://github.com/huggingface/optimum-benchmark/releases)).

I think this could be better. What do you think, @IlyasMoutawwakil?

### DIFF
--- a/.github/workflows/update_llm_perf_cuda_pytorch.yaml
+++ b/.github/workflows/update_llm_perf_cuda_pytorch.yaml
@@ -21,8 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         subset: [unquantized, bnb, awq, gptq]
-        # { name: 1xA10, runs-on: [aws-g5-4xlarge-plus] }, 
-        machine: [{ name: 1xT4, runs-on: aws-g4dn-2xlarge }]
+        # { name: 1xA10, runs-on: ... }, 
+        machine: [{ name: 1xT4, runs-on: { group: 'aws-g4dn-2xlarge' } }]
 
     runs-on: ${{ matrix.machine.runs-on }}
 
--- a/.github/workflows/update_llm_perf_cuda_pytorch.yaml
+++ b/.github/workflows/update_llm_perf_cuda_pytorch.yaml
@@ -1,12 +1,9 @@
 name: Update LLM Perf Benchmarks - CUDA PyTorch
 
-on: workflow_dispatch
-  # push:
-  #   branches:
-  #     - "add-t4-for-llm-perf-leaderboard"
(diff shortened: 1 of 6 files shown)


.github/workflows/update_llm_perf_cuda_pytorch.yaml
```diff
@@ -1,9 +1,9 @@
 name: Update LLM Perf Benchmarks - CUDA PyTorch
 
-on:
-  workflow_dispatch:
-  schedule:
-    - cron: "0 0 * * *"
+on: 
+  workflow_dispatch:  # Manual trigger
+  release:            # Trigger on new release
+    types: [published]
```
IlyasMoutawwakil, line 6: I don't think this needs commenting

baptistecolle: Ok, I can remove the comments

Good question. I think it would be more efficient to run the full benchmark with each release of the pip package, rather than on a daily basis. Running it daily seems wasteful, as the hardware remains unchanged, and we’re simply repeating the benchmark for every code change. Since users are likely to benchmark using the PyPI package, it makes more sense to align this workflow with each release. We could also run them manually if we discover any issues with our benchmarks. However, if you prefer running the benchmark daily, I can revert to that schedule. Just let me know your preference

IlyasMoutawwakil: i guess there's a misunderstanding. the daily trigger runs different benchmarks (different model+opt+quant) each time because it skips already benchmarked configurations. it is also a way to benchmark all configurations without being limited by the 6 hours time constraint of runners.

baptistecolle: Thanks for the explanation, it makes much more sense now. I removed the release schedule and left the original one

Issue #226, opened 2024-07-08
Fix quantization lib versions in llm-perf


IlyasMoutawwakil:
Following #214 we need to pin some versions for quantized models benchmarking as these libs only support a specific version of pytorch and are not failing when the kernels can't be loaded but rather falling back to slow implementations of their kernels.

Issue #214, opened 2024-05-29
Strange latency increasing on the quantized model


MartaSamoilenkoPn:
Hello!
I am testing the mistral-7b inference after quantization. I also want to test the impact of Flash Attention (sdpa, eager, fa2) on model inference. But the model decode latency is too high, and it is strange for me. I rent the virtual GPU Nvidia 4090 on Vast.ai).  I am changing the sequence length from 512 to 4096 and have the next results:
![image](https://github.com/huggingface/optimum-benchmark/assets/116800440/e0c5aca4-0d23-4655-bbd1-460ad5dff69f)
![image](https://github.com/huggingface/optimum-benchmark/assets/116800440/e8408006-8c02-4984-923c-2fefbd3e3c56)
![image](https://github.com/huggingface/optimum-benchmark/assets/116800440/d59d77ef-b154-4e5a-9acf-6765a5cc1562)

The code for the benchmarks (I change the models path and quantization/fa schemas depending on the benchmark):

```py
from optimum_benchmark.backends.pytorch.config import PyTorchConfig
from optimum_benchmark.benchmarks.inference.config import InferenceConfig
from optimum_benchmark.experiment import ExperimentConfig, launch
from optimum_benchmark.launchers.process.config import ProcessConfig
from optimum_benchmark.logging_utils import setup_logging
import torch
import os

import json


def benchmark_mistral7b(input_seqlen=1024):

    setup_logging(level="INFO")
    launcher_config = ProcessConfig(device_isolation=False)

    # batch_size = int(16384 / input_seqlen)
    batch_size = 1 

    for i in [512, 1024, 2048, 4096, 8192]:

        print("seqlen: ", i)

        benchmark_config = InferenceConfig(
            memory=True,
            latency=True,
            input_shapes={"batch_size": batch_size, "sequence_length": i},
            generate_kwargs={"max_new_tokens": 512, "min_new_tokens": 512},
        )

        backend_config = PyTorchConfig(
            model="MartaSamoilenko/Mistral-7b-int4-quantized-AWQ",
            device="cuda",
            device_ids="0",
            # torch_dtype = "float16",
            quantization_scheme="awq",
            attn_implementation = "sdpa",
            no_weights=False,
        )
        experiment_config = ExperimentConfig(
            experiment_name="mistral-awq-fa2-base",
            benchmark=benchmark_config,
            launcher=launcher_config,
            backend=backend_config,
        )
        benchmark_report = launch(experiment_config)

[...]

PR #239, opened 2024-08-01
Build from source quantization packages

baptistecolle:
Related to:
https://github.com/huggingface/optimum-benchmark/issues/226

This PR build the AutoAWQ and AUTOGPTQ from source to fix the above mentioned issue.


Makefile
```diff
@@ -23,7 +24,7 @@ build_cpu_image:
 	docker build --build-arg IMAGE=optimum-benchmark:latest-cpu --build-arg USER_ID=$(USER_ID) --build-arg GROUP_ID=$(GROUP_ID) -t optimum-benchmark:latest-cpu docker/unroot
 
 build_cuda_image:
-	docker build -t optimum-benchmark:latest-cuda docker/cuda
+	docker build --build-arg TORCH_VERSION=$(TORCH_VERSION) -t optimum-benchmark:latest-cuda docker/cuda
```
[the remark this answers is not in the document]
IlyasMoutawwakil: I think you can suppress the torch version by first installing `autoawq` (it will install whatever version it wants) then install latest torch version, then compile awq kernels.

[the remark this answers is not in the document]
baptistecolle: “AutoAWQ_kernels also requires torch==2.3.1. (https://github.com/casper-hansen/AutoAWQ_kernels/blob/main/setup.py) so this cannot work directly.

What I could do is perform a full installation with`torch==2.3.1` for AutoAWQ and AutoAWQ_kernels, then install the latest version of torch, and proceed with the rest of the package installation using the latest version of PyTorch.

However, I see this as an issue for two reasons:
1) The installation process would be a bit more convoluted. Indeed, the setup.py installation cannot handle these requirements directly (installing one version for a given package and then another version afterward). I would need to do something like `pip install . -e [autoawq]`, then `pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu121`, and finally `pip install -e .[testing,timm,diffusers,peft,auto-gptq,bitsandbytes,deepspeed]`. This process could be abstracted away in the Makefile, but it’s still not the prettiest solution.
2) I think this could cause problems later, as AutoAWQ might be incompatible with newer versions of PyTorch.

What do you think?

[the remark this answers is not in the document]
IlyasMoutawwakil: AutoAWQ kernels will definitely be incompatible, `torch` doesn't guarantee anything about extensions compatibility.
Since we're gonna build from source either ways, it's just one additional command to remove the pin from `setup.py` once it's cloned:
```bash
git clone https://github.com/casper-hansen/AutoAWQ.git
sed -i 's/torch==[0-9\.]*/torch/' AutoAWQ_kernels/setup.py
```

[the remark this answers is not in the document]
baptistecolle: Great idea! I modified it to implement your solution. I used a regex instead of sed, but the results are similar. Also, I couldn’t directly add the AutoAWQ installation to the setup.py (as it doesn’t support this use case), so I added a Python script to install the dependency

Makefile
```diff
@@ -4,6 +4,7 @@
 PWD := $(shell pwd)
 USER_ID := $(shell id -u)
 GROUP_ID := $(shell id -g)
+TORCH_VERSION := 2.3.1
```
IlyasMoutawwakil, line 7: what is it for ?

setup.py
```diff
@@ -58,12 +58,15 @@
...
+
 if USE_ROCM:
     AUTOAWQ = "autoawq@https://github.com/casper-hansen/AutoAWQ/releases/download/v0.2.1/autoawq-0.2.1+rocm571-cp310-cp310-linux_x86_64.whl"
     AUTOGPTQ = "auto-gptq@https://huggingface.github.io/autogptq-index/whl/rocm573/auto-gptq/auto_gptq-0.7.1%2Brocm5.7.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
+
 else:
-    AUTOAWQ = "autoawq==0.2.1"
-    AUTOGPTQ = "auto-gptq==0.7.1"
+    # AUTOAWQ will be installed from source via scripts/install_autoawq.py script
+    AUTOGPTQ = "auto-gptq@git+https://github.com/PanQiWei/AutoGPTQ.git"
```
IlyasMoutawwakil, line 69: we can't have this kind of links in the setup unfortunately because pypi doesn't accept them.
so yeah what I would suggest is to have normal pypi libraries in setup, and do a warning if gptq/awq kernels are not importable; as easy as trying to load the kernels directly `awq_ext`, `exl2_ext`, etc. And directing to the installation from source scripts.

baptistecolle: Unfortunately, I cannot directly check for installed packages within setup.py (without --no-build-isolation). An alternative approach could be to throw a runtime error if the required kernels are not available when attempting to use them. After the error, we could provide a warning to an installation script that users can run to install both libraries from source. What do you think of this approach?

IlyasMoutawwakil: yes that's what I meant, a runtime warning, we keep setup as clean as possible.

Makefile
baptistecolle: It was an artifact from a previous commit, I removed it

setup.py
baptistecolle: I’ve added a runtime check for the kernels and extended the CLI with new commands for installing autoAWQ and autoGPTQ from source, making it more convenient for users. The commands are optimum-benchmark +install_auto_awq_from_source=True and optimum-benchmark +install_auto_gptq_from_source=True. Unfortunately, I couldn’t make the arguments simpler (like --install_auto_awq) due to limitations with the Hydra configuration

IlyasMoutawwakil: I'm against adding this complexity to the cli command `optimum-benchmark`. let's leave them as scripts in the scripts folder, that can be used after cloning the repo (from source).

baptistecolle: Ok, I revert them back to scripts folder, so they can be run after cloning the repo. I also added this to the documentation `curl -s https://raw.githubusercontent.com/huggingface/optimum-benchmark/main/scripts/install_quantization_libs.py | python - --install-autoawq-from-source --install-autogptq-from-source` so that the user can install them more easily

Makefile
```diff
@@ -109,6 +109,7 @@ install_cli_cpu_neural_compressor:
 
 install_cli_cuda_pytorch:
 	pip install -e .[testing,timm,diffusers,peft,autoawq,auto-gptq,bitsandbytes,deepspeed]
+	curl -s https://raw.githubusercontent.com/huggingface/optimum-benchmark/main/scripts/install_quantization_libs.py | python - --install-autoawq-from-source --install-autogptq-from-source
```
IlyasMoutawwakil, line 112: where's the script in this PR I don't see it ? and why do you need to curl if the repo is cloned locally ?

baptistecolle: Sorry, I forgot to add it to my push... 😅 Also the script will only be available on main when this is merged. Right now you can test the command with `https://raw.githubusercontent.com/huggingface/optimum-benchmark/23bf7da73541bc114d4869c7c09aa7acf34025d5/scripts/install_quantization_libs.py`

The advantage is that not everyone install optimum-benchmark from source. So people with pip install optimum-benchmark can directly use our script without cloning the whole repo

IlyasMoutawwakil: > Also the script will only be available on main when this is merged

yes but it should be pushed to the OR so that I can review it 😅
> The advantage is that not everyone install optimum-benchmark from source. So people with pip install optimum-benchmark can directly use our script without cloning the whole repo
Yes but quant libs installation from source is not a "feature" of `optimum-benchmark` it's a script that we're gonna use for our own specific needs (llm-perf). if a user wants to use the script then they could either follow the installation instructions in the quant libs directly, or clone `optimum-benchmark` and use the make commands, or curl the file on their own, let's not over complicate this as it's a very super user feature (we're only responsible for the quant kernels slowdown in #214  when it impacts our benchmarks, otherwise it's on the quant lib)

baptistecolle: Ok then I will just remove the curl and adjust the documentation

baptistecolle: Done. Let me know what you think

IlyasMoutawwakil:
thanks for the addition ! it's really cool to have quant libs compiled with the latest torch, and anyone will be able to use these optimized images.

.github/workflows/test_cli_cuda_pytorch_single_gpu.yaml
```diff
@@ -49,4 +49,5 @@ jobs:
             --workdir /workspace
           run: |
             pip install -e .[testing,diffusers,timm,peft,bitsandbytes,autoawq,auto-gptq]
-            pytest -x -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not (awq)"
+            python scripts/install_quantization_libs.py --install-autoawq-from-source --install-autogptq-from-source
+            pytest -x -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not (awq)" --ignore=external_repos
```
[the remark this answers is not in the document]
baptistecolle: Good point. The problem is that the kernels are built for a specific GPU architecture. So, if we build them on an A100, they might not run on a T4 because the targeted architectures would be different. One option could be to create separate Docker images for each architecture, but this seems overly complicated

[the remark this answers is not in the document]
IlyasMoutawwakil: it doesn't have to be built for one architecture, you can use this var env to control the targeted archs https://github.com/huggingface/optimum-benchmark/blob/main/docker/cuda-ort/Dockerfile#L42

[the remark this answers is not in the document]
IlyasMoutawwakil: it will be python version (cpython) and os dependent but that's not an issue here since we use the same python version.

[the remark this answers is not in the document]
baptistecolle: Thanks for the tip. Indeed `TORCH_CUDA_ARCH_LIST` fixed the issue. How did you find out which archs to use? I used the one from https://github.com/AutoGPTQ/AutoGPTQ/blob/main/.github/workflows/build_wheels_cuda_linux.yml which is `TORCH_CUDA_ARCH_LIST="6.0 6.1 7.0 7.5 8.0 8.6 8.9 9.0+PTX"` but I am unsure which archs to target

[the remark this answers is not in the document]
IlyasMoutawwakil: i used the list from https://pytorch.org/docs/stable/generated/torch.cuda.get_arch_list.html
which has every arch that torch (specific version) was compiled for.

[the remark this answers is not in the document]
baptistecolle: Thanks, I update the list to have same TORCH_CUDA_ARCH_LIST as torch 2.4.0

.github/workflows/test_cli_cuda_pytorch_single_gpu.yaml
```diff
@@ -49,4 +49,5 @@ jobs:
             --workdir /workspace
           run: |
             pip install -e .[testing,diffusers,timm,peft,bitsandbytes,autoawq,auto-gptq]
-            pytest -x -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not (awq)"
+            python scripts/install_quantization_libs.py --install-autoawq-from-source --install-autogptq-from-source
```
IlyasMoutawwakil, line 52:
```suggestion
```

setup.py
```diff
@@ -61,9 +61,6 @@
 if USE_ROCM:
     AUTOAWQ = "autoawq@https://github.com/casper-hansen/AutoAWQ/releases/download/v0.2.1/autoawq-0.2.1+rocm571-cp310-cp310-linux_x86_64.whl"
     AUTOGPTQ = "auto-gptq@https://huggingface.github.io/autogptq-index/whl/rocm573/auto-gptq/auto_gptq-0.7.1%2Brocm5.7.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
```
IlyasMoutawwakil, line 63: you can remove this as well

IlyasMoutawwakil, line 63: we can remove these as well
```suggestion
```

setup.py
```diff
@@ -80,8 +77,8 @@
     "py-txi": ["py-txi"],
     "vllm": ["vllm"],
     # optional dependencies
-    "autoawq": [AUTOAWQ],
-    "auto-gptq": ["optimum", AUTOGPTQ],
+    "autoawq": [],
+    "auto-gptq": ["optimum"],
```
IlyasMoutawwakil, line 81: We can have our own builds in our own docker images, and have the official libs/reqs here
```suggestion
    "autoawq": ["autoawq"],
    "auto-gptq": ["optimum", "auto-gptq"],
```
if libs are already installed from source, `pip install optimum-benchmark[autoawq,auto-gptq]` normally should install nothing.

setup.py
```diff
@@ -92,6 +89,9 @@
     "peft": ["peft"],
 }
 
+if USE_ROCM:
+    EXTRAS_REQUIRE["autoawq"] = [AUTOAWQ]
+    EXTRAS_REQUIRE["auto-gptq"].append(AUTOGPTQ)
```
IlyasMoutawwakil, line 94:
```suggestion
```

scripts/install_quantization_libs.py
```diff
@@ -0,0 +1,110 @@
...
+    """Install the AutoGPTQ package from GitHub."""
+    print("Installing AutoGPTQ package.")
+    autogptq_repo_path = os.path.join(EXTERNAL_REPOS_DIR, "AutoGPTQ")
+
+    clone_or_pull_repo("https://github.com/PanQiWei/AutoGPTQ.git", autogptq_repo_path)
+
+    subprocess.run("pip install numpy gekko pandas", shell=True, check=True)
+
+    subprocess.run(
+        f"cd {autogptq_repo_path} && {sys.executable} -m pip install -vvv --no-build-isolation -e .",
```
IlyasMoutawwakil, line 44: non editable mode since we delete them afterwards
```suggestion
        f"cd {autogptq_repo_path} && {sys.executable} -m pip install -vvv --no-build-isolation .",
```

scripts/install_quantization_libs.py
```diff
@@ -0,0 +1,110 @@
...
+
+    kernels_repo_path = os.path.join(EXTERNAL_REPOS_DIR, autoawq_kernels_repo_name)
+    autoawq_repo_path = os.path.join(EXTERNAL_REPOS_DIR, autoawq_repo_name)
+
+    clone_or_pull_repo(f"https://github.com/casper-hansen/{autoawq_kernels_repo_name}", kernels_repo_path)
+
+    kernels_setup_file_path = os.path.join(kernels_repo_path, "setup.py")
+    remove_torch_from_setup(kernels_setup_file_path)
+    subprocess.run(
+        f"cd {kernels_repo_path} && {sys.executable} -m pip install --no-build-isolation -e .",
```
IlyasMoutawwakil, line 67:
```suggestion
        f"cd {kernels_repo_path} && {sys.executable} -m pip install --no-build-isolation .",
```

Makefile
```diff
@@ -158,7 +158,7 @@ test_cli_cuda_pytorch_multi_gpu:
 	pytest -s -k "cli and cuda and pytorch and (dp or ddp or device_map or deepspeed) and not awq"
 
 test_cli_cuda_pytorch_single_gpu:
-	pytest -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not awq"
+	pytest -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not awq" --ignore=external_repos
```
IlyasMoutawwakil, line 161:
```suggestion
	pytest -s -k "cli and cuda and pytorch and not (dp or ddp or device_map or deepspeed) and not awq"
```

scripts/install_quantization_libs.py
```diff
@@ -0,0 +1,110 @@
...
+        check=True,
+        env=os.environ,
+    )
+
+    clone_or_pull_repo(f"https://github.com/casper-hansen/{autoawq_repo_name}", autoawq_repo_path)
+
+    autoawq_setup_file_path = os.path.join(autoawq_repo_path, "setup.py")
+    remove_torch_from_setup(autoawq_setup_file_path)
+    subprocess.run(
+        f"cd {autoawq_repo_path} && {sys.executable} -m pip install -e .", shell=True, check=True, env=os.environ
```
IlyasMoutawwakil, line 78:
```suggestion
        f"cd {autoawq_repo_path} && {sys.executable} -m pip install .", shell=True, check=True, env=os.environ
```

PR #242, opened 2024-08-14
Fix images building

IlyasMoutawwakil:


IlyasMoutawwakil:
added an action to free some space, should work now.

PR #247, opened 2024-08-21
Add intel to leaderboard

baptistecolle:
with backends pytorch, ort and openvino


baptistecolle:
Not needed anymore, duplicate from #248

PR #248, opened 2024-08-21
Add support for intel in leaderboard

baptistecolle:


.github/workflows/update_llm_perf_intel_pytorch.yml
```diff
@@ -0,0 +1,57 @@
+name: Update LLM Perf Benchmarks - CUDA PyTorch
+
+on:
+  pull_request:
+  push:
```
IlyasMoutawwakil, line 5: avoid this

baptistecolle: yes, i know i am just doing some debugging at the moment

IlyasMoutawwakil: it triggers the workflow with the incomplete PR, which pushes to the llm perf dataset

baptistecolle: Ok thanks for the info, right now it will fails before that. I was just testing if the intel runner was picking up the job

.github/workflows/update_llm_perf_intel_pytorch.yml
```diff
@@ -0,0 +1,56 @@
+name: Update LLM Perf Benchmarks - Intel PyTorch
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
```
IlyasMoutawwakil, line 6: this has been launching the PR in the past days, hope it's intentional and didn't pollute the datasets 😅

llm_perf/hardware/update_llm_perf_intel.py
```diff
@@ -0,0 +1,219 @@
+import os
```
IlyasMoutawwakil, line 1: I don't think aggregating all intel backends in one file makes sense. each backend has specific optimizations that we might wanna explore independently, the file will only keep growing and will be very heterogenous.

llm_perf/hardware/update_llm_perf_intel.py
```diff
@@ -0,0 +1,219 @@
...
+        "4bit-awq-exllama-v2": {
+            "torch_dtype": "float16",
+            "quant_scheme": "awq",
+            "quant_config": {
+                "bits": 4,
+                "version": "exllama",
+                "exllama_config": {"version": 2, "max_input_len": 64, "max_batch_size": 1},
+            },
+        },
+    }
```
IlyasMoutawwakil, line 102: most of these configs only work with pytorch+cuda transformers.

baptistecolle: Thanks indeed i forgot to delete them

llm_perf/hardware/update_llm_perf_intel.py
baptistecolle: I modify this pr so that it is only about the pytorch backend, I will add the other backends after this one

.github/workflows/update_llm_perf_intel_pytorch.yml
baptistecolle: No, it's fine and by design. There is also a 6h timeout on the runners so having multiple scheduled runs allowed me to actually run the full benchmark

IlyasMoutawwakil:
good overall, but I would like to keep a difference between hardware and software here, the benchmarking scripts are independent from the hardware, for example the same cuda_pytorch script can run on amd rocm gpus, the cpu_pytorch can run on an intel cpu or arm cpu or amd cpu, so I don't see a need to make these scripts related to any hardware provider. on the leaderboard/frontend side yes it makes sense to indicate the provider, but no need in the backend (the same script will run on an amd cpu).

llm_perf/update_llm_perf_leaderboard.py
```diff
@@ -31,12 +34,19 @@ def gather_benchmarks(subset: str, machine: str):
...
-                gather_benchmarks(subset, machine)
-            except Exception:
-                print(f"Subset {subset} for machine {machine} not found")
+    """
+    Update the performance dataframes for all machines
+    """
+    for machine in ["1xA10", "1xA100", "1xT4", "32vCPU-C7i"]:
+        for backend in ["pytorch"]:
+            for hardware in ["cuda", "cpu"]:
+                for subset in ["unquantized", "bnb", "awq", "gptq"]:
```
baptistecolle, line 43: To prevent such a cumbersome logic, I think this could be a solution https://github.com/huggingface/optimum-benchmark/pull/258

IlyasMoutawwakil (approved):
Great !

PR #250, opened 2024-08-29
add optimum-intel ipex backend into benchmark

yao-matrix:


IlyasMoutawwakil:
very nice ! can you also add some test configs (encoders and decoders can be enough).
also a github workflow file, the formula is the same as other cpu backends.

yao-matrix:
> very nice ! can you also add some test configs (encoders and decoders can be enough). also a github workflow file, the formula is the same as other cpu backends.

thx, done and local test passed.

IlyasMoutawwakil (approved):

IlyasMoutawwakil:
Nice PR ! Merging today

PR #263, opened 2024-09-17
fix broken cuda and rocm images

baptistecolle:
AutoAWQ change its setup.py, this broke the building of our docker image

### DIFF
--- a/docker/cuda/Dockerfile
+++ b/docker/cuda/Dockerfile
@@ -35,7 +35,7 @@ RUN if [ -n "${TORCH_VERSION}" ]; then \
     # Install specific torch version if TORCH_VERSION is provided
     pip install --no-cache-dir torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
 elif [ "${TORCH_RELEASE_TYPE}" = "stable" ]; then \
-    pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
+    pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/${TORCH_CUDA} ; \
 elif [ "${TORCH_RELEASE_TYPE}" = "nightly" ]; then \
     pip install --no-cache-dir --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
 else \
--- a/docker/cuda/Dockerfile
+++ b/docker/cuda/Dockerfile
@@ -28,20 +28,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install PyTorch
 ARG TORCH_CUDA=cu124
-ARG TORCH_VERSION=stable
+ARG TORCH_RELEASE_TYPE=stable
 
-RUN if [ "${TORCH_VERSION}" = "stable" ]; then \
+RUN if [ "${TORCH_RELEASE_TYPE}" = "stable" ]; then \
     pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
-elif [ "${TORCH_VERSION}" = "nighly" ]; then \
+elif [ "${TORCH_RELEASE_TYPE}" = "nighly" ]; then \
     pip install --no-cache-dir --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
 else \
-    pip install --no-cache-dir torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
+    echo "Error: Invalid TORCH_RELEASE_TYPE. Must be 'stable' or 'nightly'." && exit 1 ; \
 fi
 
 # Install quantization libraries from source
 ENV CUDA_VERSION=12.4
 ENV TORCH_CUDA_ARCH_LIST="6.0 7.0 7.5 8.0 8.6 9.0+PTX"
-ENV CUDA_HOME=/usr/local/cuda
 
 COPY scripts/install_quantization_libs.py /internal/install_quantization_libs.py
 RUN python internal/install_quantization_libs.py --install-autogptq-from-source --install-autoawq-from-source
--- a/docker/cuda/Dockerfile
+++ b/docker/cuda/Dockerfile
@@ -29,7 +29,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Install PyTorch
 ARG TORCH_CUDA=cu124
 ARG TORCH_RELEASE_TYPE=stable
-ARG TORCH_VERSION
+ARG TORCH_VERSION=""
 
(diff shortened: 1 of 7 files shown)


docker/cuda/Dockerfile
```diff
@@ -28,14 +28,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
-RUN if [ "${TORCH_VERSION}" = "stable" ]; then \
+RUN if [ "${TORCH_RELEASE_TYPE}" = "stable" ]; then \
     pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
-elif [ "${TORCH_VERSION}" = "nighly" ]; then \
+elif [ "${TORCH_RELEASE_TYPE}" = "nighly" ]; then \
     pip install --no-cache-dir --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
 else \
-    pip install --no-cache-dir torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_CUDA} ; \
+    echo "Error: Invalid TORCH_RELEASE_TYPE. Must be 'stable' or 'nightly'." && exit 1 ; \
```
IlyasMoutawwakil, line 38: why ? what if we want to pin a specific version ?

baptistecolle: I added back the TORCH_VERSION installation. I made the change because stable is not a proper TORCH_VERSION. pip install torch==stable is not correct. This interfered with the installation of the quantization packages. I now use TORCH_RELEASE_TYPE to define if we want the latest stable version or nightly.

scripts/install_quantization_libs.py
```diff
@@ -1,29 +1,11 @@
 import argparse
 import os
-import re
 import subprocess
 import sys
 
 EXTERNAL_REPOS_DIR = "external_repos"
 
 
-def process_setup_file(setup_file_path):
```
baptistecolle, line 10: No needed anymore, as the dependency list for autoawq is now a `>=` and not a `==` so this is not needed

docker/cuda/Dockerfile
IlyasMoutawwakil: okay works, as long as it's configurable

IlyasMoutawwakil:
can you fix the same `TORCH_VERSION` issue in rocm image as well

baptistecolle:
done, rocm is fixed

IlyasMoutawwakil:
I'm not sure about that (see failing PyTorch tests), I think `https://github.com/casper-hansen/AutoAWQ/blob/main/setup.py#L10` is triggering the installation of `ipex`

docker/rocm/Dockerfile
```diff
@@ -31,16 +31,19 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-reco
...
-RUN if [ "${TORCH_VERSION}" = "stable" ]; then \
-    pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_ROCM} ; \
-elif [ "${TORCH_VERSION}" = "nightly" ]; then \
-    pip install --no-cache-dir --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/${TORCH_ROCM} ; \
-else \
+RUN if [ -n "${TORCH_VERSION}" ]; then \
+    # Install specific torch version if TORCH_VERSION is provided
     pip install --no-cache-dir torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/${TORCH_ROCM} ; \
+elif [ "${TORCH_RELEASE_TYPE}" = "stable" ]; then \
+    pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/${TORCH_ROCM} ; \
```
IlyasMoutawwakil, line 41: same thing

baptistecolle:
There are still some failing tests in our pipeline at of today:
(because of #265 - issue with the codecarbon lock)
- API CPU Tests / run_api_cpu_tests (pull_request)
- API CUDA Tests / run_api_cuda_tests (pull_request)
(issue with the use of the deprecated method - is_torch_tpu_available (we can do another PR for this i think))
- CLI CUDA Torch-ORT Multi-GPU Tests / run_cli_cuda_torch_ort_multi_gpu_tests
- CLI CUDA Torch-ORT Single-GPU Tests / run_cli_cuda_torch_ort_single_gpu_tests

scripts/install_quantization_libs.py
```diff
@@ -35,29 +18,43 @@ def clone_or_pull_repo(repo_url, repo_location_path):
...
         env=os.environ,
     )
 
+    print("Installing AutoAWQ package.")
     clone_or_pull_repo(f"https://github.com/casper-hansen/{autoawq_repo_name}", autoawq_repo_path)
-    autoawq_setup_file_path = os.path.join(autoawq_repo_path, "setup.py")
-    process_setup_file(autoawq_setup_file_path)
+    if os.environ.get("IMAGE_FLAVOR") in ["cuda", "rocm"]:
+        autoawq_setup_file_path = os.path.join(autoawq_repo_path, "setup.py")
+        process_setup_file_for_autoawq(autoawq_setup_file_path)
```
IlyasMoutawwakil, line 57: we only use these scripts in these two images, what are we checking here ?

baptistecolle: I just realized i forgot to add cuda-ort. But the idea is to patch the setup.py to remove the installation of ipex for the none cpu based image https://github.com/casper-hansen/AutoAWQ/blob/main/setup.py#L10. Just to confirm we want to have ipex for the cpu image or is better not to have it (as we could have a amd cpu)?

baptistecolle: it is true it is redundant, i will remove it

docker/cuda/Dockerfile
```diff
@@ -28,14 +28,18 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Install PyTorch
 ARG TORCH_CUDA=cu124
-ARG TORCH_VERSION=stable
+ARG TORCH_RELEASE_TYPE=stable
+ARG TORCH_VERSION
```
IlyasMoutawwakil, line 32: docker will keep warning us for not setting a default value here
```suggestion
ARG TORCH_VERSION=""
```

docker/rocm/Dockerfile
```diff
@@ -26,21 +26,24 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-reco
...
     update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1 && \
-    pip install --no-cache-dir --upgrade pip setuptools wheel requests && \
+    pip install --no-cache-dir --upgrade pip setuptools wheel requests pyrsmi && \
     cd /opt/rocm/share/amd_smi && pip install .
 
 # Install PyTorch
 ARG TORCH_ROCM=rocm5.7
-ARG TORCH_VERSION=stable
+ARG TORCH_RELEASE_TYPE=stable
+ARG TORCH_VERSION
```
IlyasMoutawwakil, line 35:
```suggestion
ARG TORCH_VERSION=""
```

Issue #260, opened 2024-09-12
Error: Another instance of codecarbon is already running


j-irion:
Hello, 
When I try to track the energy consumption with the benchmark, I get the following error:
`Error: Another instance of codecarbon is already running. Turn off the other instance to be able to run this one. Exiting.`

Afterwards I get another Error: 
`AttributeError: 'EmissionsTracker' object has no attribute '_scheduler'`

I noticed that, when I reboot the machine, the first run crashes later than the consequent runs and also saves to the task_codecarbon.json. Here are the logs of the first and second run:
[first_run_after_reboot.log](https://github.com/user-attachments/files/16982005/first_run_after_reboot.log)
[second_run.log](https://github.com/user-attachments/files/16982006/second_run.log)

I tried running the benchmark with the following config:

```yaml
defaults:
  - benchmark
  - scenario: inference
  - launcher: process
  - backend: pytorch
  - _base_
  - _self_

name: pytorch_llama_8b_energy

launcher:
  device_isolation: true
  device_isolation_action: warn

backend:
  model: meta-llama/Meta-Llama-3.1-8B
  device: cuda
  device_ids: 0,1
  device_map: auto
  torch_dtype: float16

scenario:
  memory: true
  latency: true
  energy: true

  warmup_runs: 10
  iterations: 10
  duration: 10

  input_shapes:
    batch_size: 2
    sequence_length: 512
  generate_kwargs:
    max_new_tokens: 32
    min_new_tokens: 32
```

IlyasMoutawwakil:
Hi, it must be the new release of codecarbon that's breaking stuff https://pypi.org/project/codecarbon/#history, I'll see if it can be fixed as part of the #169

j-irion:
Thanks for your effort! Do you know, by any chance, which was the last compatible version of codecarbon?

IlyasMoutawwakil:
2.6 probably since, our tests were passing then.

PR #265, opened 2024-09-18
Fix issue with CodeCarbon lock

regisss:
From CodeCarbon v2.7, a lock file is introduced (see [here](https://github.com/mlco2/codecarbon/blob/v2.7.0/codecarbon/emissions_tracker.py#L638)) to check whether another instance of codecarbon is running. If yes, an error is raised. One has to call `my_energy_tracker.stop()` to release the lock file.

This PR adds a `stop` method in the `EnergyTracker` class that should be called once the energy tracker is not needed anymore.
It also updates an import from `huggingface_hub` since v0.25 introduced a change (the new import is backward-compatible).

Fixes https://github.com/huggingface/optimum-benchmark/issues/260.

### DIFF
--- a/optimum_benchmark/hub_utils.py
+++ b/optimum_benchmark/hub_utils.py
@@ -9,7 +9,7 @@ from typing import Any, Dict, Optional
 import pandas as pd
 from flatten_dict import flatten, unflatten
 from huggingface_hub import create_repo, hf_hub_download, upload_file
-from huggingface_hub.utils._errors import HfHubHTTPError
+from huggingface_hub.utils import HfHubHTTPError
 from typing_extensions import Self
 
 LOGGER = getLogger("hub_utils")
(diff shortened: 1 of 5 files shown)


IlyasMoutawwakil:
the failing tests are from the ongoing changes in #263

IlyasMoutawwakil:
Acquiring the lock and releasing it between processes will probably not work (or cause issues) in multi-gpu/process setting (`torchrun`). I can see one process waiting for the other to release the lock to start measuring and causing async issues. We actually don't tests that (codecarbon+multi-gpu). I think a better thing to do here is to set `allow_multiple_runs` to True.
https://github.com/mlco2/codecarbon/blob/v2.7.0/codecarbon/emissions_tracker.py#L239-L255

regisss:
> Acquiring the lock and releasing it between processes will probably not work (or cause issues) in multi-gpu/process setting (`torchrun`). I can see one process waiting for the other to release the lock to start measuring and causing async issues. We actually don't test that (codecarbon+multi-gpu). I think a better thing to do here is to set `allow_multiple_runs` to True. https://github.com/mlco2/codecarbon/blob/v2.7.0/codecarbon/emissions_tracker.py#L239-L255

True for multi-process this is needed. I wouldn't set `allow_multiple_runs` to True for single-process benchmark though, we don't really know how several trackers can interfere with each other.

regisss:
> the failing tests are from the ongoing changes in #263

Let's wait for #263 to be merged then

IlyasMoutawwakil:
> we don't really know how several trackers can interfere with each other
Tbh it would be much better if we have one tracker in distributed settings

PR #272, opened 2024-09-23
ipex backend enhancements

yao-matrix:
1. add feature-extraction task mapping for ipex backend, to support embedding models benchmark
2. change examples' no_weight to false, no_weight will allocate weight buffers and random initialize them, which will ruin performance in numa cases, 2x perf drop for decoding phases

### DIFF
--- a/examples/ipex_bert.yaml
+++ b/examples/ipex_bert.yaml
@@ -23,7 +23,7 @@ scenario:
 
 backend:
   device: cpu
-  no_weights: true
+  no_weights: false
   export: true
   torch_dtype: bfloat16
   model: bert-base-uncased
--- a/examples/ipex_llama.yaml
+++ b/examples/ipex_llama.yaml
@@ -32,6 +32,6 @@ scenario:
 backend:
   device: cpu
   export: true
-  no_weights: true
+  no_weights: false
   torch_dtype: bfloat16
   model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
--- a/optimum_benchmark/backends/ipex/utils.py
+++ b/optimum_benchmark/backends/ipex/utils.py
@@ -1,6 +1,7 @@
 TASKS_TO_IPEXMODEL = {
     "fill-mask": "optimum.intel.IPEXModelForMaskedLM",
     "text-generation": "optimum.intel.IPEXModelForCausalLM",
+    "feature-extraction": "optimum.intel.IPEXModel",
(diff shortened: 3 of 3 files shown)


yao-matrix:
@IlyasMoutawwakil, pls help review, thx.

IlyasMoutawwakil:
Hi ! are you sure about this `no_weight will allocate weight buffers and random initialize them`, no_weights only interferes with random generators used inside the model, so instead of using these methods
https://github.com/huggingface/optimum-benchmark/blob/01e4e599381bbd166de56057d1649aba9e7d2100/optimum_benchmark/backends/transformers_utils.py#L190-L205
It'll use the fasterst one of them https://github.com/huggingface/optimum-benchmark/blob/01e4e599381bbd166de56057d1649aba9e7d2100/optimum_benchmark/backends/transformers_utils.py#L207-L208
How does that ruin performance ?

yao-matrix:
> Hi ! are you sure about this `no_weight will allocate weight buffers and random initialize them`, no_weights only interferes with random generators used inside the model, so instead of using these methods
> 
> https://github.com/huggingface/optimum-benchmark/blob/01e4e599381bbd166de56057d1649aba9e7d2100/optimum_benchmark/backends/transformers_utils.py#L190-L205
> 
> 
> It'll use the fasterst one of them
> https://github.com/huggingface/optimum-benchmark/blob/01e4e599381bbd166de56057d1649aba9e7d2100/optimum_benchmark/backends/transformers_utils.py#L207-L208
> 
> 
> How does that ruin performance ?

seems random init funtions are mostly single thread function(e.g. [here](https://github.com/pytorch/pytorch/blob/44c871c34bda37c409457e9a2afa8ed7ffd75502/torch/nn/init.py#L27)), and numa memory allocation strategy somewhat follows a allocate while first write way. So in random initialization case, the weights memory will allocate near to the core executing the random initialization logic, which, for example, is done all in numa domain 0. So when model forward computation happens, which may spread across numa domain 0 and 1, the compute in numa domain 1 must far fetch the data from numa domain 0(since the data already allocated while random initialization), which bring far memory access cost. 

data evidence, using GCP c4-standard-96 instance to run `meta-llama/Meta-Llama-3-8B` w/ bs=1, in_seq=256, out_seq=64, 
no_weights true: decoding throughput 16.37
no_weights false: decoding throughput 8.69

IlyasMoutawwakil:
very interesting behavior ! thanks for investigating it

PR #279, opened 2024-09-25
Pass backend name to EnergyTracker in Training scenario

asesorov:
Seems there is a missing required `backend` parameter in `EnergyTracker` initialization in training scenario, which causes a value error.


IlyasMoutawwakil:
Thanks, can you add `energy=not rocm_system()` here https://github.com/huggingface/optimum-benchmark/blob/main/tests/test_api.py#L67

PR #287, opened 2024-10-09
add testing for example folder

baptistecolle:


IlyasMoutawwakil:
I prefer these examples to be distributed across the same CIs that already exist, for the simple reason that these CIs already cover the entirety of device+backend combinations.

examples/ipex_bert.yaml
```diff
@@ -25,5 +25,4 @@ backend:
   device: cpu
   no_weights: false
   export: true
-  torch_dtype: bfloat16
```
baptistecolle, line 28: removed because `BF16 weight prepack needs the cpu support avx_ne_convert or avx512bw, avx512vl and avx512dg`. We are running the test on `ubuntu-latest`. This must be some old CPU so it is not the best to test ipex IMO. Do you have an idea or preference for a more modern intel CPU to use for the CI?

.github/workflows/test_cli_cpu_openvino.yaml
```diff
@@ -47,5 +48,10 @@ jobs:
           pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
           pip install -e .[testing,openvino,diffusers,timm]
 
-      - name: Run tests
-        run: pytest tests/test_cli.py -s -k "cli and cpu and openvino"
+      # - name: Run cli tests
+      #   run: |
+      #     pytest tests/test_cli.py -s -k "cli and cpu and openvino"
+
+      - name: Run tests from example folder
```
baptistecolle, line 55: This takes forever because the CPU for testing `ubuntu-latest` is not the most optimal for the CI, I think. Maybe we should also update the machine that run those tests

examples/tei_bge.yaml
```diff
@@ -3,14 +3,11 @@ defaults:
   - scenario: inference
   - launcher: inline
   - backend: py-txi
+  - _base_
   - _self_
 
 name: tei_bert
 
-launcher:
```
baptistecolle, line 10: Device isolation is only supported on CUDA or ROCM

examples/tei_bge.yaml
```diff
@@ -3,14 +3,11 @@ defaults:
...
-  device_isolation: true
-  device_isolation_action: warn
-
```
baptistecolle, line 13: Device isolation is only supported on CUDA or ROCM

examples/trt_llama.yaml
```diff
@@ -15,7 +15,6 @@ launcher:
 backend:
   device: cuda
   device_ids: 0
-  no_weights: true
```
baptistecolle, line 18: no weights does not seem to be supported by the trt backend

baptistecolle:
I now use the CLI workflows to also run the different examples, however I believe this should lead to some refactoring. I think it would be make sense to maybe rename the CLI tests. For example `CLI CUDA Pytorch Tests` should be `CUDA Pytorch Tests` and `test_cli_cuda_pytorch.yaml` should now be `test_cuda_pytorch.yaml`. Wdyt?

baptistecolle:
Also, for your information, there are also a lot of device+backend combinations that do not have an example config: 
- cpu+llama_cpp -> the example is on a mps device
- cpu+pytorch
- cuda+onnxruntime
- cuda+torchort
- rocm+pytorch

optimum_benchmark/backends/vllm/backend.py
```diff
@@ -124,7 +124,6 @@ def batch_offline_engine_generate(self, inputs: Dict[str, Any], kwargs: Dict[str
                     n=kwargs.get("num_return_sequences"),
                     max_tokens=kwargs.get("max_new_tokens"),
                     min_tokens=kwargs.get("min_new_tokens"),
-                    use_beam_search=kwargs.get("num_beams") > 1,
```
baptistecolle, line 127: use_beam_search is not a parameter of samplingParams https://github.com/vllm-project/vllm/blob/e9d517f27673ec8736c026f2311d3c250d5f9061/vllm/sampling_params.py#L105

tests/test_examples.py
```diff
@@ -0,0 +1,216 @@
...
+CUDA_PYTORCH_SCRIPTS = [
+    "pytorch_bert.py",
+    "pytorch_gpt2.py",
+]
+
+# Those tests are not run on the CI/CD pipeline as they are currently broken
+UNTESTED_YAML_CONFIGS = [
+    "energy_star.yaml",
+    "trt_llama.yaml",
+]
```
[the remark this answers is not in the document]
IlyasMoutawwakil: energy star is under work, but trt should work fine

optimum_benchmark/backends/vllm/backend.py
IlyasMoutawwakil: yeh it was removed in the latest version https://github.com/vllm-project/vllm/releases/tag/v0.6.2

examples/ipex_bert.yaml
IlyasMoutawwakil: I prefer testing on a more advanced cpu machine, but minimal in terms of size / vCPUs.

examples/trt_llama.yaml
IlyasMoutawwakil: it used to be, I don't remember when it was removed tbh.

optimum_benchmark/scenarios/energy_star/scenario.py
```diff
@@ -7,7 +7,7 @@
 from tqdm import tqdm
```
IlyasMoutawwakil, line 7: better not change anything in energy star, the energy start PR will take care of it, otherwise will have conflicts

optimum_benchmark/backends/tensorrt_llm/config.py
```diff
@@ -24,7 +24,7 @@ class TRTLLMConfig(BackendConfig):
     world_size: int = 1
     gpus_per_node: int = 1
 
-    max_prompt_length: int = 128
+    max_prompt_length: int = 256
```
IlyasMoutawwakil, line 27: any specific reason ?

PR #292, opened 2024-10-21
[DRAFT] Set vllm version limit to avoid upstream API incompatibility

vicoooo26:
According to vllm upstream v0.6.2 [release note](https://github.com/vllm-project/vllm/releases/tag/v0.6.2), beam search has been soft deprecated and attribute `use_beam_search` in [SamplingParams](https://github.com/vllm-project/vllm/blob/v0.6.3/vllm/sampling_params.py#L173) has been removed in v0.6.3. To keep [vllm backend](https://github.com/huggingface/optimum-benchmark/blob/main/optimum_benchmark/backends/vllm/backend.py#L127 ) available without refactoring, setting vllm max version limit will help.

### DIFF
--- a/optimum_benchmark/backends/vllm/backend.py
+++ b/optimum_benchmark/backends/vllm/backend.py
@@ -117,35 +117,31 @@ class VLLMBackend(Backend[VLLMConfig]):
             self.pretrained_model.add_request(
                 inputs=prompt,
                 request_id=str(i),
-                params=SamplingParams(
-                    ignore_eos=True,
-                    detokenize=True,
-                    seed=self.config.seed,
-                    n=kwargs.get("num_return_sequences"),
-                    max_tokens=kwargs.get("max_new_tokens"),
-                    min_tokens=kwargs.get("min_new_tokens"),
-                    use_beam_search=kwargs.get("num_beams") > 1,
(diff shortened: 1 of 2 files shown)


IlyasMoutawwakil:
Hi, I don't think limiting the version and all the new models/fixes that come with it is justifiable for something like a deprecated argument. I prefer compatibility with the latest version because that's more "fair" when it comes to benchmarks.

vicoooo26:
Hi, I do agree with you about 'fairness'. To make it compatible with the latest vllm, maybe we could leverage the log probabilities and replace `use_beam_search` just like vllm [beam search implementation](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/llm.py#L415), or we could use some higher API instead of vllm core. What do you think?

vicoooo26:
Hi, I remove `use_beam_search` param and let `logprobs = 2 * num_beams` if a user passes `num_beams` that greater than 1. If it looks good to you, I will create a new PR.

vicoooo26:
I have tested it already.

vicoooo26:
Tested it in the dev environment and

vicoooo26:
as

vicoooo26:
asasas

PR #293, opened 2024-10-24
Feat: reimplement vllm backend beam search using logprobs

vicoooo26:
Vllm upstream changed `beam search` implementation. Corresponding updates should be made to get compatibility with the latest vllm.
Reimplement vllm backend beam search using logprobs according to huggingface [transformers implementation](https://github.com/huggingface/transformers/blob/v4.45.2/src/transformers/generation/beam_search.py#L534)

### DIFF
--- a/optimum_benchmark/backends/vllm/backend.py
+++ b/optimum_benchmark/backends/vllm/backend.py
@@ -117,35 +117,33 @@ class VLLMBackend(Backend[VLLMConfig]):
             self.pretrained_model.add_request(
                 inputs=prompt,
                 request_id=str(i),
-                params=SamplingParams(
-                    ignore_eos=True,
-                    detokenize=True,
-                    seed=self.config.seed,
-                    n=kwargs.get("num_return_sequences"),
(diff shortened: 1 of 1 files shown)


optimum_benchmark/backends/vllm/backend.py
```diff
@@ -117,35 +117,33 @@ def batch_offline_engine_generate(self, inputs: Dict[str, Any], kwargs: Dict[str
...
+            seed=self.config.seed,
+            n=kwargs.get("num_return_sequences"),
+            max_tokens=kwargs.get("max_new_tokens"),
+            min_tokens=kwargs.get("min_new_tokens"),
+            logits_processors=kwargs.get("logits_processors", None),
+        )
+        # following huggingface transformers implementation
+        # https://github.com/huggingface/transformers/blob/v4.45.2/src/transformers/generation/beam_search.py#L534
+        if kwargs.get("num_beams") > 1:
+            params.logprobs = 2 * kwargs.get("num_beams")
```
IlyasMoutawwakil, line 139:
```suggestion

```

